PR #6, opened 2024-03-11
V1.6 2024 03 11 run

kaitejohnson:
Updated codebase for 2024-03-11 run plus 2024-03-11 forecasts


_targets.R
```diff
@@ -132,7 +132,7 @@ list(
     command = write_config(
       save_config = TRUE,
       location = "US",
-      prod_run = TRUE,
+      prod_run = FALSE,
```
dylanhmorris: Change intended?

_targets.R
```diff
@@ -152,7 +152,7 @@ list(
     command = write_config(
       save_config = TRUE,
       location = NULL,
-      prod_run = TRUE,
+      prod_run = FALSE,
```
dylanhmorris: Change intended?

_targets.R
kaitejohnson:
```suggestion
      prod_run = TRUE,
```

_targets.R
kaitejohnson:
```suggestion
      prod_run = TRUE,
```

_targets.R
```diff
@@ -112,7 +112,7 @@ list(
...
```
kaitejohnson:
```suggestion
      prod_run = TRUE,
```

kaitejohnson (commented):
Suggested reverting back to prod = TRUE

dylanhmorris (approved):
LGTM

PR #87, opened 2024-06-24
Implementation of NCP with normal approximation to iid sum of gammas

kaitejohnson:
Starting over here with NCP and rescaled to relative shedding variability * mean genomes shed (where mean genomes shed is estimated in log scale). 

Additionally, I took Adrian's suggestion and am now using modified versions of his `lognormal3` functions which parameterize the lognormal with the log of the mean (not the $mu$!) and the scale parameter in unit scale,  so that $MC(t)$ is the mean not the median and `sigma` is on the unit scale. 

Able to get only a few divergences with a relative diffuse prior on the `cv` in individual shedding, so this seems like progress! Going to test out running this on all locations for our comparison forecast date to see how it performs... 

Borrowed from https://github.com/adrian-lison/EpiSewer


dylanhmorris:
Closing as not planned.

Issue #51, opened 2024-06-23
Mechanistic handling of observation error in ww scaling with contributing infections


kaitejohnson:
## Problem
The expected variability in the wastewater measurements is a function of the number of contributing infections. The model doesn't currently have any component to account for the population size/number of infections 

Assuming we can write the infected individuals distribution of number of genomes shed over the course of infection as:
$$g_i \sim \mathrm{Gamma}(\kappa, \theta)$$

Which has mean $\kappa\theta$.

Then the sum of the genomes in each infected individual should be:
```math
 \sum_{i=1}^{I(t)} g_{it}\sim \mathrm{Gamma}(\kappa I(t) , \; \theta)$
```
with mean $I(t)\kappa\theta$, because the mean of a gamma is shape *scale

The original write-up in #57 has the scale parameter also scaling with $I(t)$. 

Context:
https://github.com/cdcent/cfa-forecast-renewal-ww/issues/57

PR #88, opened 2024-06-24
NCP gamma sum approximation 2nd pass

kaitejohnson:
Attempt to address #51 
So I think I got something working that uses the normal approximation to the sum of gammas that is used in EpiSewer. 
This does a rescaling of the number of genomes shed per infection to separate the "relative shedding intensities" of each individual and the mean number of genomes shed per infection $\mu_g$. 
Thus the total number of genomes shed in the sewershed is:

```math
V(t) = \sum_{t=0}^{\tau}\mu_g \zeta(t-\tau)s(\tau)
```
where $\zeta(t)$ is the sum of iid Gamma distributed individual relative shedding intensities with mean 1 and coefficient of variation $cv$:
```math
\zeta(t) = \sum_{i=1}^{i= I(t)} Gamma(\alpha = \frac{1}{cv^2}, \beta =  \frac{1}{cv^2}) =  Gamma(\alpha = \frac{I(t)}{cv^2}, \beta =  \frac{1}{cv^2})
```
This is implemented using the non-centered parameterization of a normal approximation as is described in[ EpiSewer ](https://github.com/adrian-lison/EpiSewer/blob/f0664050ad0643915e6987ac439f92c0a797b4b9/inst/stan/functions/gamma2.stan#L29). We estimate a matrix of standard normals and multiply by the standard deviation of the sum of gammas (mean of I(t) stdev of sqrt(I(t))*cv)

The coefficient of variation in individual shedding is the key parameter here. If we use the very tight and informative prior (imposing very low variability in fecal shedding) of `cv ~halfN(0.1, 0.025)`, we don't have major convergence issues (but we also know this is basically the same as using the no variation model). 

If we set the prior to be uninformative of `cv~ halfN(0.5, 0.5)` we get some convergence issues but also some variation in the estimate of the cv.


dylanhmorris:
Closing as not planned.

PR #89, opened 2024-06-24
WIP exp gamma implementation

kaitejohnson:
Another attempt at #51 
This does a rescaling of the number of genomes shed per infection to separate the "relative shedding intensities" of each individual and the mean number of genomes shed per infection $\mu_g$. 
Thus the total number of genomes shed in the sewershed is:

```math
V(t) = \sum_{t=0}^{\tau}\mu_g \zeta(t-\tau)s(\tau)
```
where $\zeta(t)$ is the sum of iid Gamma distributed individual relative shedding intensities with mean 1 and coefficient of variation $cv$:
```math
\zeta(t) = \sum_{i=1}^{i= I(t)} Gamma(\alpha = \frac{1}{cv^2}, \beta =  \frac{1}{cv^2}) =  Gamma(\alpha = \frac{I(t)}{cv^2}, \beta =  \frac{1}{cv^2})
```

Then we implement this by sampling directly in the model block what we refer to in the code as `zeta_bar`, which returns the log of the output of the gamma draw, and then we log transform the mean number of genomes and the calculation of the expected concentration in the transformed parameters block. 
```
 for (i in 1:n_subpops){
          for (j in 1:(uot + ot + ht)){
            zeta_bar[i,j] ~ expgamma(i_site_t[i,j]/(cv^2), 1/((cv^2))); // Sum of iid gamma individual relative shedding intensities
          }
        }
```

However, this seems to struggle a lot more to sample parameter space when we set an uninformative prior on the cv.


dylanhmorris:
Closing as not planned.

PR #95, opened 2024-06-27
Point workflows at action endpoints, remove custom/vendored actions

### LINKED ISSUE #94: Replace vendored github actions
Now that development is in public, the following vendored/custom actions can be replaced with workflows that point to maintained actions elsewhere on github:
- [x] `pre-commit`
- [x]  installing `cmdstan`

dylanhmorris:
Closes #94


kaitejohnson (approved):
I never really understood what all this was doing, but this definitely looks simpler

PR #97, opened 2024-07-05
change call in eval config list to refer to path not dir

kaitejohnson:
Found a typo in the `write_eval_config.R` script with `dir` vs `path`. Then realized that we needed to also add the same styler patch as in https://github.com/CDCgov/ww-inference-model/pull/2/files


src/write_eval_config.R
```diff
@@ -128,7 +128,7 @@ write_eval_config <- function(locations, forecast_dates,
     baseline_score_table_dir = baseline_score_table_dir,
     output_dir = output_dir,
     hub_subdir = hub_subdir,
-    retro_rt_dir = retro_rt_dir,
+    retro_rt_path = retro_rt_path,
```
damonbayer, line 131: Can you explain this change? The code has many references to `retro_rt_path`, but none to `retro_rt_dir`. https://github.com/search?q=repo%3ACDCgov%2Fwastewater-informed-covid-forecasting%20retro_rt&type=code

wweval/R/format_data_for_stan.R
```diff
@@ -2,7 +2,8 @@
 #'
 #' @param model_type string indicating which model we are getting data for
 #' Options are `ww` or `hosp`
-#' @param forecast_date string indicating the forecast date
+#' @param forecast_date string indicating the forecast date in ISO8 convention
```
damonbayer:
```suggestion
#' @param forecast_date string indicating the forecast date in ISO 8601 convention
```

I think? https://en.wikipedia.org/wiki/ISO_8601

damonbayer (commented):
Question and a comment

src/write_eval_config.R
kaitejohnson: Right, that's why I am changing it to `retro_rt_path`. This is pointing to a full path where the Rt draws (including the file and its extension) will be, so it should be called a path. We tried to call everything a `dir` if it points to a folder.

wweval/R/format_data_for_stan.R
kaitejohnson: Good catch, probably need to fix this in the package code

src/write_eval_config.R
damonbayer: Not sure why I didn't understand that yesterday 🙃

Thanks!

damonbayer (approved):

PR #98, opened 2024-07-10
Hot fix retro rt formatting

kaitejohnson:
Minor change to allow for allow for more flexibility in the way that retrospective R(t)s are formatted (initially the column name was `state_abbr`, now its `state_abb`)


dylanhmorris (approved):
LGTM

PR #99, opened 2024-07-11
Quick fix of the hospital admissions post processing

kaitejohnson:
Very happy we unlinked post processing and fitting, bc this will need to be rerun on azure but just the postprocessing. The change here is I just want to have the full set of quantiles for the hospital admissions model (not just the nowcast and forecast period).

I also scope creep fixed a bug in the calling of a function argument in the targets pipeline.

Perhaps can rerun this as part of the azure demo tomorrow...


dylanhmorris (approved):
LGTM

PR #100, opened 2024-07-11
Figure 2 rough draft components

kaitejohnson:
The point of this figure is to show an example comparing the forecasts of the wastewater model and the hospital admissions model for a few states (can decide on the states later), paired with the fits/forecasts to a subset of the wastewater data for that state. 

This PR creates (separately) the ggplot components for the example hospital admissions forecasts colored by model type + the fits and forecasts to the wastewater concentrations for those example states. 

See private repo for produced [figures](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/759).

Note: The goal here is just to have the structure in place to generate the figures. All formatting decisions will be made in a second round of revisions. Am still using the `{targets}` pipeline for creating all of these, which enables caching and allows us to pass the post-processed objects/ mix analysis in with figure production when needed.


scratch/temp_ms_figs.R
dylanhmorris, line 1: Did you mean to keep both this and the `wweval/manuscript_figs.R` file? If so, why?

_targets_eval_postprocessing.R
```diff
@@ -388,6 +388,45 @@ head_to_head_targets <- list(
...
+      date_to_plot = "2024-01-15"
+    )
+  ),
+  tar_target(
+    name = fig2_ct_3,
+    command = make_fig2_ct(
+      all_ww_quantiles_sq,
+      loc_to_plot = "WA",
+      date_to_plot = "2024-01-15"
+    )
```
dylanhmorris, line 459: Having the concentration plots be one ggplot per location while the forecast plots are a single faceted ggplot might make axis alignment, etc down the road more difficult. 

Related: as written, this produces different plot aspects depending on the number of wastewater sites within a location. If you're going to do each location individually, you'll probably want functionality to add null "site" facets for locations with fewer than the maximum number of plotted sites, so that the x axes can be aligned.

dylanhmorris (commented):
Looks largely good. A couple minor comments.

scratch/temp_ms_figs.R
kaitejohnson: Oh no I can remove this

_targets_eval_postprocessing.R
kaitejohnson: True, my thinking was whenever we can to use faceting, I just didn't do it for the wastewater sites because of the different number of sites thing. 

I think of this as a bit of a downstream formatting question e.g. do we want to show a set number of sites even if the state has a different number of sites, or do we want to always show all sites? 

Defer to you here on how we want to do this in the short term -- can either make them all separate plots or set up faceting now adding null "site" facets and or enforcing a set number of sites to be displayed

kaitejohnson:
Converting back to a draft because I realized I forgot to make the second half of what was planned for Fig 2 (the forecast comparison plots by location and horizon, with CRPS underlaid)

kaitejohnson:
Ok @dylanhmorris this is ready for review. Let me know if you'd prefer I just put all the figures in one PR, I figured breaking it up would be slightly easier. All the drafts are in the link to the cdcent repo issue. 

In order to make the completed planned figure 2, we'd have to repeat a bunch of generations of the subpanels for the different locations, I left that out of the targets pipeline for now.

wweval/R/manuscript_figs.R
```diff
@@ -0,0 +1,423 @@
...
+
+#' Make hospital forecast comparison figure
+#'
+#' @param hosp_quantiles A tibble containing the calibrated hospital admissions
+#' data, the evaluation hospital admissions data, and the quantiles of the
+#' calibrated and forecasted admissions
+#' @param loc_to_plot A  string indicating the state abbreviations of the state
+#' to plot
+#' @param horizon_to_plot A string indicating what horizon period to plot,
+#' either `nowcast`, `1 wk`, or `4 wks`
```
dylanhmorris:
```suggestion
#' one of `nowcast`, `1 wk`, or `4 wks`
```

wweval/R/manuscript_figs.R
```diff
@@ -0,0 +1,423 @@
...
+}
+
+#' Make CRPS underlay figure
+#'
+#' @param scores A tibble of scores by location, forecast date, date and model,
+#' the ouput of `scoringutils::score()` on samples.
+#' @param loc_to_plot A  string indicating the state abbreviations of the state
+#' to plot
+#' @param horizon_to_plot A string indicating what horizon period to plot,
+#' either `nowcast`, `1 wk`, or `4 wks`
```
dylanhmorris:
```suggestion
#' one of `nowcast`, `1 wk`, or `4 wks`
```

wweval/R/manuscript_figs.R
```diff
@@ -0,0 +1,423 @@
...
+                                   horizon_to_plot) {
+  scores_by_horizon <- scores |>
+    dplyr::filter(location == loc_to_plot) |>
+    dplyr::mutate(
+      horizon = dplyr::case_when(
+        date <= forecast_date ~ "nowcast",
+        date > forecast_date & date <= forecast_date +
+          lubridate::days(7) ~ "1 wk",
+        date > forecast_date + lubridate::days(21) &
+          date <= forecast_date + lubridate::days(28) ~ "4 wks"
```
dylanhmorris: For D.R.Y. reasons I think this should be broken out into its own function. Given that the results will be used for filtering later on, would be good to have a `TRUE ~ NA` at the end for "none of the above`.

A couple other questions:
1. are you sure you want string horizons instead of, say, an integer `horizon_weeks` (or for that mattter string/factor horizons that _derive_ from an integer `horizon_weeks`?)
2. For safety, would be good to specify a minimum date for the `"nowcast"` if possible.

dylanhmorris (changes requested):
A couple minor but key questions @kaitejohnson

wweval/R/manuscript_figs.R
kaitejohnson: Good call, I also repeat this a bunch of times. Will make these changes tomorrow! 

1. I think that makes the most sense, bc in Fig 3 we have one that is continuous, and this also makes it easier to set the order properly. The only thing is that the nowcast is 10 days so this will be -2 and -1 but the -2 will be incomplete. I think its fine though since we will mostly be analyzing the outputs using this categorically. 
2. Agreed-- though as this will apply to both scores and quantiles we might just have to do something where you pass in the maximum number of days of the nowcast instead of using the data (since the calibration data will be present in the quantiles, but not in the scores)

wweval/R/add_horizons.R
dylanhmorris, line 1: Is anything in these functions specific to scores vs. quantiles? Why not have just one function?

kaitejohnson:
@dylanhmorris Let me know what you think of this. I added the additional `horizon_weeks` and `horizon` columns upstream when creating the long form of the quantiles and the scores, and then pass them in to the figure functions. I think this will avoid redundancy. 

I still have an open question about what `horizon_week` to assigned the forecast date (see my teams post). Once we've reached a consensus will edit that.

dylanhmorris (changes requested):

wweval/R/add_horizons.R
kaitejohnson: Yeah the only thing that is is the quantiles are for the calibration, nowcast, and forecast period whereas the scores are only for the nowcast/forecast period. There's an extra `period` column in the quantiles, and we would want to set the categorical `horizon` as `calibration` when `period == "calibration"` (I think)

kaitejohnson: I could probably write a switch for it though...

kaitejohnson:
Update, going to switch to creating `horizon_days` folder and then creating categorical `horizon` columns from that. Going to lump the forecast date into the nowcast, since in theory, you could have partially observed data on the day you are making a forecast

kaitejohnson:
Ok this is ready for re-review, let me know if you want me to make them into one function though @dylanhmorris

wweval/R/add_horizons.R
```diff
@@ -0,0 +1,73 @@
...
+#' @export
+add_horizons_to_quantiles <- function(quantiles) {
+  quantiles_w_horizons <- quantiles |>
+    dplyr::mutate(
+      horizon_days = as.numeric(date - forecast_date)
+    ) |>
+    dplyr::mutate(
+      horizon = dplyr::case_when(
+        period == "calibration" ~ "calibration",
+        period != "calibration" & horizon_days < 0 ~ "nowcast",
```
dylanhmorris: if kept as-is, this needs to change to <= 0
```suggestion
        period != "calibration" & horizon_days <= 0 ~ "nowcast",
```

wweval/R/add_horizons.R
dylanhmorris, line 1: Given this redesign, I think we should make it one function that always takes the `max_days_nowcast` argument. If you want to be consistent with `period != "calibration"`, just use that to compute the value of `max_days_nowcast` you pass to _both_. Related: I wouldn't give `max_days_nowcast` a default value.

dylanhmorris (changes requested):

_targets_eval_postprocessing.R
```diff
@@ -301,6 +301,15 @@ head_to_head_targets <- list(
   # Get the full set of quantiles, filtered down to only states and
   # forecast dates with sufficient wastewater for both ww model and hosp only
   # model. Then join the convergence df
+  tar_target(
+    name = max_nowcast_days,
+    command = as.numeric(
+      max(all_ww_hosp_quantiles$forecast_date) -
+        max(all_ww_hosp_quantiles$date[!is.na(
+          all_ww_hosp_quantiles$calib_data
+        )])
```
dylanhmorris: I'm not convinced this is the quantity you want to be computing. Can you explain how you're trying to set the `max_nowcast_days` here? As the set of days before the forecast date without _any_ admissions data on the date the forecast was made? Do you want it a different value for each `forecast_date`/`scenario`/`location` trio?

kaitejohnson: Here I am just finding a single value which represents the number of days between the forecast date and the last day of observed data. I am just using the maximum value to find this so I can get a single number. In practice, this is the same for all `forecast_date`/`scenario`/`locations`, data ends on a Friday and the forecast is made the following Monday. 

In theory, this could be different for each `forecast_date`/`scenario`/`location` , which of course means you couldn't just pass it as a function argument but would need to calculate it based on the `calib_data` within the `df` (which is what I was originally doing with the `add_horizons_to_quantiles()` function.

I am not following how there would be a way to have `max_nowcast_days` passed as an argument to `add_horizons()` without passing a single scalar and assuming it is the same for all combinations. This is why I broke the functions up originally, because with the quantiles, we have `calib_data` and with the scores, I have already filtered to remove days with observed data (we don't score those days).

I probably made this more complicated than it needed to be!

kaitejohnson: @dylanhmorris lmk what you think, when this is done I'll need to port the changes to my fig3 branch

wweval/R/add_horizons.R
```diff
@@ -41,26 +7,26 @@ add_horizons_to_quantiles <- function(quantiles) {
...
+                         max_nowcast_days) {
+  df_w_horizons <- df |>
     dplyr::mutate(
       horizon_days = as.numeric(date - forecast_date)
     ) |>
     dplyr::mutate(
       horizon = dplyr::case_when(
-        horizon_days <= -max_days_nowcast ~ "calibration",
-        horizon_days > -max_days_nowcast & horizon_days <= 0 ~ "nowcast",
+        horizon_days <= -max_nowcast_days ~ "calibration",
```
dylanhmorris:
```suggestion
        horizon_days <= -!!max_nowcast_days ~ "calibration",
```

wweval/R/add_horizons.R
```diff
@@ -41,26 +7,26 @@ add_horizons_to_quantiles <- function(quantiles) {
...
+        horizon_days > -max_nowcast_days & horizon_days <= 0 ~ "nowcast",
```
dylanhmorris:
```suggestion
        horizon_days > -!!max_nowcast_days & horizon_days <= 0 ~ "nowcast",
```

_targets_eval_postprocessing.R
dylanhmorris: Thanks, this clarifies things. If `max_nowcast_days` is a vector of the same length as the dataframe, the comparisons will work, as `case_when` is vectorized.

That said, if your goal is to start the "nowcast" on the period on the first day without observed hospitalizations, the simplest thing to do is to have a "`last_hosp_data_date` column in both all dataframes that get passed to `add_horizons()` and compare directly to that. Is that easy to do? If so, we should do it here. If not, we can stick with this approach.

kaitejohnson: I think adding that column to both dataframes is the most elegant solution! I can just add a target that creates that mapping of `location`/`forecast_date`/`last_hosp_data_date` and then left join it to both dataframes

dylanhmorris: Sounds good. Ping me when ready for review.  Minor note: since every entry in the `score_df` and `quantile_df` should have an entry in the `last_data_date_df` (not committed to these names, just for clarity), I would do `right_join(last_data_date_df, score_df, unmatched = "error")` or `left_join(score_df, last_data_date_df, unmatched = "error")`. See https://dplyr.tidyverse.org/reference/mutate-joins.html

kaitejohnson: `last_data_date_df` will contain a larger subset of combinations of `location` and `forecast_date` than `scores_filtered` and `hosp_quantiles_filtered`, bc both `score_df` and `quantile_df` will have been filtered to only include locations and forecast dates where the wastewater is sufficient and the model has converged. 

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/b0d50364e996b8ffedf2f7f1e7797efa1b56e8df/_targets_eval_postprocessing.R#L354

So if we flag it to error if unmatched, the "extra" mappings will cause an error.

dylanhmorris: Apologies, I was wrong about the behavior of that argument. In that case, just make sure to do a join that removes the non-matches, i.e.  either `right_join(last_data_date_df, score_df)` or `left_join(score_df, last_data_date_df)`

wweval/R/add_horizons.R
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+  return(df_w_horizons)
+}
+
+#' Get a map of the location, forecast date, and last hospital admissions data
+#' date
+#'
+#' @param df A tibble containing the following columns: `forecast_date`,
+#' `location`,`date`, `calib_data`
```
dylanhmorris:
```suggestion
#' `location`,`date`, `calib_data`.
#' `calib_data` should be `NA` for any dates for which there was 
#'  not hospital admissions data to fit to. 
```

dylanhmorris (approved):
LGTM. One documentation wording suggestion.

dylanhmorris (approved):

PR #101, opened 2024-07-11
add functionality to save cmdstan csvs for both models

kaitejohnson:
@damonbayer As requested, here is functionality to save the raw cmdstan csvs. Feel free to change the directory of where to save them.

Not sure if we want to write a wrapper function around this that defaults to not saving (if someone runs this many times, these won't over write bc they are by default time-stamped).


damonbayer (approved):
Thanks! This is really useful to me, but I don't know if it needs to be part of the vignette for everyone to see. Up to you.

kaitejohnson:
> Thanks! This is really useful to me, but I don't know if it needs to be part of the vignette for everyone to see. Up to you.

Agreed, probably not! Going to close

PR #103, opened 2024-07-15
add a warning message and a new file to write out missing files

kaitejohnson:
This PR adds the functionality to generate warning messages when our "roll-your-own" version of `tar_combine()` (called `combine_outputs()` tries to load in a file and fails. This is helpful because some of these seem to have been killed jobs on Azure that we would only know by spot checking manually (which is not ideal) and we would just need to rerun them.

I added functionality to write it into a file, we could make this optional. My thinking here is most of this analysis is really for our own evaluation analysis rather than very general tooling, so its maybe okay to do something like this, but happy to change if we want to push towards maximizing user generality.

### DIFF
--- a/wweval/R/combine_outputs.R
+++ b/wweval/R/combine_outputs.R
@@ -68,6 +68,10 @@ combine_outputs <- function(output_type =
         error = function(e) {}
       )
     } else {
+      warning(glue::glue(
+        "File missing for {this_scenario}",
+        "in {this_location} on {this_forecast_date}"
+      ))
       # Create a tibble of the combos that are missing, to save
       this_failed_output <- tibble(
         scenario = this_scenario,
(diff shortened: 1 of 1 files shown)


wweval/R/combine_outputs.R
```diff
@@ -66,11 +67,41 @@ combine_outputs <- function(output_type =
...
 
   if (nrow(combined_output) == 0) {
     combined_output <- NULL
   }
+
+  if (nrow(flag_failed_output) != 0) {
+    warning(glue::glue(
+      "File missing for {this_scenario}",
+      "in {this_location} on {this_forecast_date}"
+    ))
```
dylanhmorris: I believe that, as written, this will always warn about the final scenario/location/date processed, versus warning about any individual flagged trios. To fix this, the actual warning should happen in the for-loop, whereas the table of files missing can be written out at the end.

dylanhmorris (changes requested):

dylanhmorris (approved):
Would be nice to have a unit test for this function to confirm it does indeed create the file as needed, but that can be a separate PR. Make an issue?

Issue #84, opened 2024-06-24
Make first draft of evaluation manuscript figures


kaitejohnson:
## Goal
The `_targets_eval_postprocessing.R` should contain all the outputs needed to generate the proposed figures currently

## Requirements
- [ ] First draft of the figures, even though the model outputs might still change. Can just generate each sub plot individually for now, we can deal with ggarrange or mucking them up for formatting later

PR #108, opened 2024-07-19
Fig 5 components

kaitejohnson:
This partially completes #84, adding the components from the Hub comparison figure. 
The drafted figures are available[ here](
image
). Remaining issues with formatting are:
- need to order the x-axis of the distribution plots so preferably Oct-Mar is on the left and Feb-Mar is on the right
- something off with the quartiles and not automatically being labeled/assigned in order
- I think because we only have 7 models right now these plots have weird little ridges. Could consider adding more models


kaitejohnson:
Closing this, doing everything in #112

PR #110, opened 2024-07-22
Add epidemic phase figure as a check for how its working

kaitejohnson:
This PR adds an additional figure (that we might want to reformat if we were to really put in the supplement) that I am using to check how the epidemic phase classification is working. As expected, I don't think this is exactly what we want because it is telling us about the trend in infection dynamics currently, which is about 2 weeks on average earlier than what we see in the hospital admissions signal. See fig[ here](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/759#issuecomment-2242774911).

Not for this PR necessarily, but we should probably have a discussion about changing this algorithm a bit @zsusswein tagging you for awareness/thoughts since you suspected this would be an issue...


kaitejohnson:
Closing this bc its in the #112

PR #112, opened 2024-07-23
Make drafts of patchwork figures

### LINKED ISSUE #106: Use patchwork to draft all figures and discuss formatting
## Goal
#84's goal is to produce code to generate all the subplots in the figures. This issue the goal is to use `patchwork` in R to format the ggplot objects for discussion among the team

## Requirements
- [ ] drafts of combined plots for manuscript

kaitejohnson:
This PR closes #106. Still a WIP of patching all the proposed figures together, will post drafts [here](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/761)


wweval/R/plots.R
```diff
@@ -1051,7 +1051,16 @@ get_plot_hub_performance <- function(all_scores,
...
-    dplyr::filter(model != {{ baseline_model }})
+    dplyr::filter(model != {{ baseline_model }}) |>
+    dplyr::mutate(
+      fig_order =
+        dplyr::case_when(
+          period == "Feb 2024-Mar 2024" ~ 1,
+          period == "Oct 2023-Mar 2024" ~ 0
+        ),
+      horizon = forcats::fct_reorder(horizon, fig_order)
+    )
```
dylanhmorris: I assume you meant to order `period` here, not `horizon`, yes? Another case I'd handle with base `factor`:
```suggestion
    dplyr::mutate(
      period = factor(period, ordered = TRUE, levels = c(
          "Feb 2024-Mar 2024", "Oct 2023-Mar 2024"))
    )
```

wweval/R/plotting_style.R
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+#' Get plot components (colors for now)
+#'
+#' @return a list of model colors and horizon colors to be passed to
+#' `scale_fill_manual()` and `scale_color_manual()`
+#' @export
+plot_components <- function() {
+  pal_horizons <- RColorBrewer::brewer.pal(12, "Paired")
```
dylanhmorris, line 83: Looks like `RColorBrewer` needs to be added to `DESCRIPTION` under `Imports`

wweval/R/plotting_style.R
```diff
@@ -0,0 +1,115 @@
...
+    "UT-Osiris" = pal_model[6],
+    "COVIDhub-baseline" = pal_model[7]
+  )
+
+  colors_list <- list(
+    horizon_colors = horizon_colors,
+    model_colors = model_colors
+  )
+  return(colors_list)
+}
```
dylanhmorris, line 116: I'd actually create a `scale_color_model()`, `scale_color_horizon()` etc from your pre-computed `horizon_colors`, e.g.

```R
scale_color_horizon <- function(...) {
   return (scale_colour_manual(values = unlist(horizon_colors), ...))
}
```

Then you can do:

```R
my_plot <- my_plot + scale_color_horizon() + other_ggplot_stuff()
```

dylanhmorris (commented):
A few comments while you work, to avoid a mega-review later @kaitejohnson

wweval/R/plots.R
kaitejohnson: Whoops good catch yes!

wweval/R/plotting_style.R
kaitejohnson: I would still need to pass in the `colors` though right? So this would just avoid using `values = colors$horizon_colors` and instead allow you to do
`scale_color_horizon(colors)`?

dylanhmorris: no, in this approach the `scale_fill_` function would "know" the colors. Concrete example from some paper code of mine where I defined a color scale for variants:

First I defined the variant-name-to-color mapping as a `list`:

https://github.com/dylanhmorris/aerosol-stability-voc/blob/2eb4181bead1d89ebeeba173c2c824c1e1a6c925/varstab/R/plotting_style.R#L221-L236


Then I used that `list` to define custom `scale_fill_variant()` and `scale_color_variant()` functions:

https://github.com/dylanhmorris/aerosol-stability-voc/blob/2eb4181bead1d89ebeeba173c2c824c1e1a6c925/varstab/R/plotting_style.R#L239-L271

dylanhmorris: If I'd also needed the particular values of that list to be accessible in other project code via a function call, I could have defined another function, e.g. something like this:

````R
get_variant_color <- function(variant_name) {
    if(!(variant_name %in% names(variant_colors)) {
        cli::cli_abort("No color defined for variant {variant_name}")
   }
   return (variant_colors[[variant_name]])
}

Comment (kaitejohnson) [reply wweval/R/plotting_style.R:116]: @dylanhmorris Those links aren't working for me for some reason. 

So I made those lists in `plot_components()` but I think it just isn't clear to me how to make that accessible in other project code without simply calling the function (which is what I have in here).

**Comment by kaitejohnson** [comment]
@dylanhmorris This is also ready for review. I see this very much as a draft, and imagine we will have lots of new issues to make from this

File: wweval/R/plotting_style.R (hunk 3 above)
Comment (dylanhmorris) [reply wweval/R/plotting_style.R:116]: Oops. Try the links now. Should explain a bit more what I mean. In the linked example, the list mapping values to colors exists at script scope in the file where the functions that use it are defined (but it is not exported).

File: wweval/R/get_table_sufficient_ww.R [hunk 4]
```diff
@@ -9,8 +9,8 @@
 #' sufficient to inform a forecast, based on the critera that was used for
 #' the Hub submissions.
 #'
-#' @param ww_quantiles The dataframe of the quantiled ww predictions with the
-#' real data alongside it (labeled as calib data)
+#' @param path_to_ww_vintaged_data A character string indicating the path
+#' to where only the vintaged wastewater data are saved
 #' @param delay_thres The maximum number of days of delay between the last
 #' wastewater data point and the forecat date, before we would flag a state as
```
dylanhmorris: Typo 
```suggestion
#' wastewater data point and the forecast date, before we would flag a state as
```

wweval/DESCRIPTION
```diff
@@ -72,7 +72,11 @@ Imports:
     tibble,
     cfaforecastrenewalww,
     cmdstanr (>= 0.7.1),
-    forcats
+    forcats,
```
dylanhmorris, line 75:
```suggestion
```
We've determined that `forecats` isn't needed, yes?

wweval/R/get_table_sufficient_ww.R
```diff
@@ -9,8 +9,8 @@
 #' sufficient to inform a forecast, based on the critera that was used for
 #' the Hub submissions.
 #'
-#' @param ww_quantiles The dataframe of the quantiled ww predictions with the
-#' real data alongside it (labeled as calib data)
+#' @param path_to_ww_vintaged_data A character string indicating the path
+#' to where only the vintaged wastewater data are saved
```
```suggestion
#' to where vintaged wastewater data are saved
```

wweval/R/get_table_sufficient_ww.R
dylanhmorris, line 1: Can you explain what's going on here? Seems like the calculation you're doing requires only the input data. Why not load that directly? It seems as though here the input data is getting added to and then extracted back from the output.

wweval/R/plotting_style.R
dylanhmorris: this is now done.

wweval/R/plots.R
```diff
@@ -1051,7 +1051,13 @@ get_plot_hub_performance <- function(all_scores,
       "location"
     )) |>
     dplyr::mutate(relative_wis = interval_score / baseline_score) |>
-    dplyr::filter(model != {{ baseline_model }})
+    dplyr::filter(model != {{ baseline_model }}) |>
+    dplyr::mutate(
+      period = factor(period, ordered = TRUE, levels = c(
+        "Feb 2024-Mar 2024", "Oct 2023-Mar 2024"
+      ))
+    )
```
dylanhmorris: Probably best not to hard-code this here within the plotting function if it can be avoided, versus writing a function to order `period`s elsewhere.

wweval/R/plots.R
dylanhmorris, line 1: Not for this PR, but it would be good to split this file up as well.

wweval/R/ms_fig4.R
```diff
@@ -262,71 +333,80 @@ make_fig4_rel_crps_overall <- function(scores,
...
+        "model"
+      )) |>
+      dplyr::mutate(horizon = "overall")
+  } else {
+    scores_by_horizon <- scores
+    scores_overall <- scores |>
+      dplyr::mutate(
+        horizon = "overall"
+      )
+  }
```
dylanhmorris: When / how do you use this switch?

wweval/R/ms_fig4.R
```diff
@@ -439,34 +537,149 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
+      x_axis_title_size = 8,
+      y_axis_title_size = 8
+    ) +
+    scale_fill_brewer(palette = "Set3")
+
 
   return(p)
 }
+
+#' Make average WIS over time for model comparison
```
dylanhmorris:
```suggestion
#' Plot average CRPS over time for model comparison
```

wweval/R/ms_fig5.R
```diff
@@ -0,0 +1,465 @@
...
+    dplyr::mutate(relative_wis = interval_score / baseline_score) |>
+    dplyr::filter(model != {{ baseline_model }}) |>
+    dplyr::mutate(
+      fig_order =
+        dplyr::case_when(
+          period == "Feb 2024-Mar 2024" ~ 1,
+          period == "Oct 2023-Mar 2024" ~ 0
+        ),
+      horizon = forcats::fct_reorder(period, fig_order)
+    )
```
dylanhmorris: As above, should handle this with an `order_period` function that uses base `factor`, not `forecats::fct_reorder()`

dylanhmorris (commented):
Some comments / questions @kaitejohnson

wweval/R/plotting_style.R
kaitejohnson: Ok, I see -- so in this example you are just creating the mapping of variants to colors, then writing a function that uses that mapping by default, rather than including the mapping within the function itself.

If you'd prefer, I can rewrite the `plot_components()` to do this for the things that I have set to have specific colors (which are models, horizons, and phases).

wweval/R/plots.R
kaitejohnson: Ok, these are not manuscript plots but just ones that we had made throughout the pipeline

wweval/DESCRIPTION
kaitejohnson: Still needed here:
https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/23cdf0644b0dec7237b18c0265be49a22b349d12/wweval/R/ms_fig5.R#L380

I think I tried the factor order thing and it didn't work here because its internal to the dataframe rather than an external order we tell it.

wweval/R/ms_fig4.R
kaitejohnson: Originally, we had summarized across the days in the horizon for each point. We changed this to keep each day's score. But this allows us to revert back to summarizing across the horizon easily. The switch is in https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/cf970e8ac72422eb69be6b02bb348ac4f57abc88/wweval/R/ms_fig4.R#L22

wweval/R/get_table_sufficient_ww.R
kaitejohnson: The input data is not a part of the postprocessing pipeline!

kaitejohnson: This is only a function of the outputs from `post_process` which are scores, hospital quantiles, and ww quantiles.

wweval/R/plotting_style.R
kaitejohnson: Let me know if I am understanding the ask correctly!

wweval/DESCRIPTION
dylanhmorris: Thanks. There are indeed ways to do this without `forecats`, but this is the exact use case for `forecats::fct_reorder`, so I'm happy to add the dependency

wweval/R/plotting_style.R
dylanhmorris: Resolved f2f with @katiejohnson that for now we will call `scale_color_manual` etc on the output of `plot_components()` in the figure code, and defer creating custom scales.

wweval/R/get_table_sufficient_ww.R
dylanhmorris: Resolved f2f that we will address this join/re-extract issue in a separate PR.

wweval/R/add_horizons.R
dylanhmorris, line 1: Not for this PR, but we should maybe split all the ordering functions off into their own file, and also possibly have them error if the `df` passed does not contain the column that is to be ordered.

wweval/R/add_horizons.R
```diff
@@ -84,3 +85,22 @@ order_horizons <- function(df) {
 
   return(df_w_order)
 }
+#' Get the order of the time periods for plotting
+#'
+#' @param df a dataframe containing the column name `period` which will
+#' contain character strings indicating what period the scores correspond to
+#'
+#' @return a dataframe containing the same columns as `df` but now ordered
+#' in terms of the specified order.
```
dylanhmorris:
```suggestion
#' @return a dataframe containing the same columns as `df` but with the `period`
#' factor now ordered so that `"Oct 2023-Mar 2024"` comes first and 
#' `"Feb 2024-Mar 2024"` comes second.

```

wweval/R/add_horizons.R
```diff
@@ -64,17 +64,18 @@ get_last_hosp_data_date_map <- function(df) {
 #' contain character strings indicating the horizon (by week, nowcast period,
 #' or overall) of the score/quantile/sample.
 #'
-#' @return a dataframe containing the same columns as `df` but with the
-#' addition of `fig_order` and reordered in terms of `fig_order` for plotting.
+#' @return a dataframe containing the same columns as `df` but reordered
+#' in terms of horizon order
```
dylanhmorris:
```suggestion
#' @return a dataframe containing the same columns as `df` with the `horizon`
#' column now an ordered factor with ordered levels as follows:
#' `"overall", "calibration", "nowcast", "1 wk", "2 wks", "3 wks", "4 wks"`
```

wweval/R/plotting_style.R
```diff
@@ -0,0 +1,116 @@
...
+    "ww" = pal_model[1],
+    "hosp" = pal_model[2],
+    "cfa-wwrenewal(retro)" = pal_model[1],
+    "cfa-wwrenewal(real-time)" = pal_model[5],
+    "cfa-hosponlyrenewal(retro)" = pal_model[2],
+    "COVIDhub-4_week_ensemble" = pastel_model[3],
+    "UMass-trends_ensemble" = pastel_model[4],
+    "UT-Osiris" = pastel_model[5],
+    "COVIDhub-baseline" = pastel_model[6]
+  )
```
dylanhmorris, line 109: Note that you may need to change this depending on #119

wweval/R/ms_fig5.R
dylanhmorris:
```suggestion
    order_period()
```

dylanhmorris: @kaitejohnson looks like this still needs to change.

wweval/R/plots.R
dylanhmorris:
```suggestion
    order_period()
```

dylanhmorris: looks like this one still needs to change @kaitejohnson. Have made a suggestion.

wweval/R/ms_fig4.R
dylanhmorris: I think we're unlikely to go back and if we do we may well want a separate panel creation function (or want two functions that call the same internal helper after summarizing things differently). So I think you can delete the switch.

dylanhmorris (changes requested):
Very close. A few minor things still need resolving.

wweval/R/plots.R
kaitejohnson: Hmm are you sure, mine is showing this as using `order_periods()`

wweval/R/ms_fig5.R
kaitejohnson: This is done!

wweval/R/add_horizons.R
kaitejohnson: https://github.com/CDCgov/wastewater-informed-covid-forecasting/issues/121

wweval/R/plotting_style.R
kaitejohnson: Yep!

wweval/R/ms_fig5.R
kaitejohnson: It is changed https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/b30ff226d1e296930ed9a7f9cbabd723b5357cc6/wweval/R/ms_fig5.R#L189

wweval/R/plots.R
kaitejohnson: https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/b30ff226d1e296930ed9a7f9cbabd723b5357cc6/wweval/R/plots.R#L1055
Changed

kaitejohnson:
Ok @dylanhmorris this is ready for re-review pending passing CI. Will let you resolve conversations, note that the `order_periods()` ones have been updated.

wweval/R/plots.R
dylanhmorris: I see it now, thanks!

dylanhmorris (approved):
LGTM, thanks @kaitejohnson!

PR #117, opened 2024-07-29
Add targets to save a pdf of the wastewater data by state

### LINKED ISSUE #116: Visual inspection of wastewater data across the season
## Goal
Want to see if there are periods of state's wastewater data that we should exclude bc of poor data quality

## Context
We encountered some of these during the real-time submissions. There are likely more. First pass is just to look at the final dataset and see if we flag any states/periods of time with wastewater data that we should exclude

kaitejohnson:
This PR closes #116. It adds to the pipeline to save the wastewater data visualized with one page per state as a pdf. See PDF [here](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/762). We will use this to visually inspect whether there are states/dates where wastewater data should be excluded. In particular, I have identified a few states to look at from last week based on very poor model performance over all.

### DIFF
--- a/_targets_eval_postprocessing.R
+++ b/_targets_eval_postprocessing.R
@@ -106,6 +106,33 @@ upstream_targets <- list(
       ww_data_mapping = eval_config$ww_data_mapping
     )
   ),
+  tar_target(
+    name = grouped_eval_ww_data,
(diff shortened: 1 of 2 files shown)


dylanhmorris (approved):
LGTM

PR #118, opened 2024-07-29
Convert table of exclusions to the one needed for our workflow

### LINKED ISSUE #115: Add hospital data exclusions to pre-processing pipeline rerun
## Goal
@zsusswein sent a table of hospital admissions exclusions from the past season. We want to add these to azure, and incorporate them into the run. First we will test this locally with just a few states/dates with known exclusions, then set up for a model rerun following exclusions of outlier/flawed wastewater data

## Requirements
- [ ] test that incorporating admissions exclusions works as expected locally.

kaitejohnson:
This PR closes #115. Currently just working on converting from one format to the way our model expects it to be formatted which is just by location, forecast date, and date to exclude. Might need to just make this manually though

### DIFF
--- a/scratch/convert_to_table_of_exclusions.R
+++ b/scratch/convert_to_table_of_exclusions.R
@@ -1,0 +1,22 @@
+# Load in the raw table from NNH
+raw_table <- arrow::read_parquet(file.path(
+  "input",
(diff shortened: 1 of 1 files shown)


kaitejohnson:
Closing this as there is no single table of exclusion history

PR #119, opened 2024-07-29
add other models from covidhub, remove those that are incomplete

### LINKED ISSUE #114: Add all models that submitted to the forecast hub each week to hub comparison analysis
## Goal
Add all the relevant COVID hub models to the analysis in Figure 5 where we compare our real-time and retrospective model performance to that of the other Hub models

## Context
We prototyped the Hub analysis with just a few individual team's submitted forecasts. We should go through and identify the teams that submitted every week and put all of them into the analysis (even though we will only show results compared to say the best performing one and the ensemble in the main text). This might be helpful. https://viz.covid19forecasthub.org/?as_of=2024-02-03&interval=95%25&target_var=hosp&xaxis_range=2023-11-17&xaxis_range=2024-05-13&yaxis_range=-79.38888888888891&yaxis_range=2488.3888888888887&model=COVIDhub-ensemble&model=cfa-wwrenewal&location=12

## Requirements
- [ ] score all models that submitted to the Hub each week
- [ ] redo analysis with completed set of scores

kaitejohnson:
This PR closes #114. It just expands the Hub models that we score (I had prototyped on a smaller set). These are all the ones highlighted for comparison [here](https://viz.covid19forecasthub.org/?as_of=2024-02-03&interval=95%25&target_var=hosp&xaxis_range=2023-11-17&xaxis_range=2024-05-13&yaxis_range=-79.38888888888891&yaxis_range=2488.3888888888887&model=COVIDhub-ensemble&model=cfa-wwrenewal&location=12)). However, there are a number of models (6 teams) that don't have forecasts for every date. We can continue to exclude them, or we could do something like allow one or two missing weeks.


src/write_eval_config.R
```diff
@@ -72,8 +72,17 @@ write_eval_config <- function(locations, forecast_dates,
   retro_rt_path <- file.path("input", "retro_Rt", "Rt_draws.parquet")
   score_subdir <- file.path("output", "eval", "hub")
   hub_model_names <- c(
```
dylanhmorris: If the goal is to use all of the models submitted to the hub for the period of interest (and then filter for completeness), ideally we should query the hub itself for that information rather than hard coding it.

kaitejohnson: Do we have a better way of querying than loading in the data from github? I talked to @ekr-cfa ad @k88hudson and they had said that they just pull from github. I thought the hubverse stuff was still in development, Happy to be proven wrong. But otherwise, I don't think we want to pull in all of it because theres a ton and only certain teams have submitted recently

kaitejohnson:
Should probably use the `zoltr` API https://reichlab.io/zoltr/

wweval/R/score.R
```diff
@@ -186,6 +186,88 @@ make_baseline_score_table <- function(all_ww_scores,
...
+  # Submit query, poll job, get job data
+
+  forecast_data <- zoltr::do_zoltar_query(
+    zoltar_connection = zoltar_connection,
+    project_url = project_url,
+    query_type = "forecasts",
+    models = NULL, # all models by default
+    units = c("06"), # We could query all of them, but this will be very slow.
+    # Suggest we use a single state as a proxy. I don't know what this refers
+    # to since the state codes here don't correspond to those in github...
```
dylanhmorris: It shouldn't be that slow if we're just doing a small number of forecast dates? Ideally I think we'd do something like "at least X% submissions for at least Y% of locations"

If the codes are mostly numeric I'd guess they're the 2-digit state FIPS codes that flusight uses. See https://github.com/cdcent/cfa-forecasttools/blob/main/data-raw/flusight_location_table.R

kaitejohnson: It's these which make no sense to me, clearly a few numbers that look like they could be 2-digit state FIPS codes but not the complete list, and then these 5 digit ones. I can just leave it as NULL and then subset for the ones included in the ensemble. My patience was low so can see how long it actually takes with all of these and the 22 forecast dates. 
```
valid names=['US', '01', '06', '11', '17', '21', '25', '30', '34', '37', '42', '46', '50', '55', '66', '74', '01041', '01051', '01061', '01071', '01077', '01087', '01097', '01107', '01115', '01123', '01133', '02020', '02070', '02110', '02150', '02170', '02180', '02195', '02240', '02290', '04011', '04019', '04027', '05009', '05019', '05027', '05035', '05047', '05057', '05063', '05073', '05083', '05093', '05101', '05111', '05121', '05129', '05139', '05145', '06005', '06013', '06021', '06031', '06041', '06049', '06059', '06067', '06073', '06081', '06087', '06099', '06109', '06111', '08005', '08013', '08019', '08031', '08041', '08051', '08059', '08069', '08077', '08087', '08097', '08105', '08113', '08123', '09005', '09013', '10005', '12005', '12015', '12027', '12037', '12047', '12055', '12063', '12071', '12081', '12087', '12097', '12105', '12113', '1212'
````

Comment (kaitejohnson) [reply wweval/R/score.R]: Looks like you can pass in all the 2 digit ones and it returns them all, my bad I should have tried first. I can check for the proportion of all locations as well.

Comment (kaitejohnson) [reply wweval/R/score.R]: somehow it is still querying even with only our 22 forecast dates in my test file...

Comment (kaitejohnson) [reply wweval/R/score.R]: Hmm yeah as written this took over an hour then had a timeout error. See change I made to only grab the 28 day ahead target, and add a criteria for proportion of locations

Comment (dylanhmorris) [reply wweval/R/score.R]: Nice idea on only grabbing one target per forecast/location pair.

File: scratch/scratch_hub_api.R [hunk 3]
```diff
@@ -0,0 +1,114 @@
...
+  unique(eval_config$location_hosp)
+)
+
+# Submit query, poll job, get job data
+
+forecast_data <- do_zoltar_query(
+  zoltar_connection = zoltar_connection,
+  project_url = project_url,
+  query_type = "forecasts",
+  models = NULL, # "cfa-wwrenewal",
```
dylanhmorris:
```suggestion
  models = NULL,
```

_targets_eval_postprocessing.R
```diff
@@ -1207,6 +1227,13 @@ hub_comparison_plots <- list(
...
+      "CEPH-Rtrend_covid",
+      "MOBS-GLEAM_COVID",
+      "CMU-TimeSeries",
+      "SGroup-RandomForest",
+      "UMass-gbq",
+      "UMass-sarix",
+      "UT-Osiris",
       "cfa-wwrenewal(real-time)",
       "cfa-wwrenewal(retro)",
       "cfa-hosponlyrenewal(retro)"
```
dylanhmorris, line 1236: Long run we will presumably want this to be generated programmatically, no? Fine for now.

src/write_eval_config.R
```diff
@@ -73,10 +73,10 @@ write_eval_config <- function(locations, forecast_dates,
...
   retro_rt_path <- file.path("input", "retro_Rt", "Rt_draws.parquet")
   score_subdir <- file.path("output", "eval", "hub")
-  hub_model_names <- c(
-    "COVIDhub-4_week_ensemble", "UMass-trends_ensemble",
-    "UT-Osiris", "COVIDhub-baseline"
-  )
+  # Proportion of forecast dates that a model must have submitted for to be
+  # included in the Hub analysis
+  prop_dates_for_incl_hub <- 19 / 22
+  prop_locs_for_incl_hub <- 51 / 52
```
dylanhmorris: I might drop this lower?

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+                                    project_name = "COVID-19 Forecasts") {
+  # get state abbreviation codes
+  state_codes <- cfaforecastrenewalww::loc_abbr_to_flusight_code(
+    unique(locations)
+  )
+
+  if (prop_dates_for_incl_hub > 1) {
+    cli::cli_abort(c(
+      "Proportion of forecast dates required for hub inclusion",
+      "must be less than 1."
```
dylanhmorris:
```suggestion
      "must be greater than 0 and less than or equal to 1."
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+#' @description
+#' This function uses the `zoltr` R package to connect to the Zoltar database
+#' which contains forecasts from the COVID Hub forecast project, and query it
+#' for the specified forecast dates. It queries for all dates, computes the
+#' proportion of dates for which the model has submitted, filters for models
+#' that have submitted for greater than the specified proportion of forecast
+#' dates for inclusion, and returns the vector of model names.
+#'
+#'
+#' @param prop_dates_for_incl_hub Numeric less than 1 indicating the inclusion
```
dylanhmorris:
```suggestion
#' @param prop_dates_for_incl_hub Numeric greater than 0 and less than or equal to 1 indicating the inclusion
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+#' threshold for the proportion of forecast dates that a model must have
+#' submitted forecasts to be included in analysis
+#' @param prop_locs_incl_hub Numeric less than 1 indicating the inclusion
```
dylanhmorris:
```suggestion
#' @param prop_locs_incl_hub Numeric greater than 0 and less than or equal to 1 indicating the inclusion
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+                                    prop_locs_for_incl_hub,
+                                    forecast_dates,
+                                    locations,
+                                    project_name = "COVID-19 Forecasts") {
+  # get state abbreviation codes
+  state_codes <- cfaforecastrenewalww::loc_abbr_to_flusight_code(
+    unique(locations)
+  )
+
+  if (prop_dates_for_incl_hub > 1) {
```
dylanhmorris:
```suggestion
  if (prop_dates_for_incl_hub > 1 | prop_dates_for_incl_hub <= 0) {
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    cli::cli_abort(c(
+      "Proportion of forecast dates required for hub inclusion",
+      "must be less than 1."
+    ))
+  }
+
+  if (prop_locs_for_incl_hub > 1) {
```
dylanhmorris:
```suggestion
  if (prop_locs_for_incl_hub > 1 | prop_locs_for_incl_hub <= 0) {
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    cli::cli_abort(c(
+      "Proportion of locations required for hub inclusion",
+      "must be less than 1."
```
dylanhmorris:
```suggestion
      "must be greater than 0 and less than or equal to 1."
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+  # Submit query, poll job, get job data
+
+  forecast_data <- zoltr::do_zoltar_query(
+    zoltar_connection = zoltar_connection,
+    project_url = project_url,
+    query_type = "forecasts",
+    models = NULL, # all models by default
+    units = state_codes,
+    # We could query all of them, but this was very slow. This ensures
+    # that the forecasts submitted have at leastr eached 28 days.
```
dylanhmorris:
```suggestion
    # that the forecasts submitted have at least reached 28 days.
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    dplyr::distinct(timezero) |>
+    dplyr::pull() |>
+    length()
+
+  n_forecasts_per_model <- forecast_data |>
+    dplyr::distinct(timezero, model) |>
+    dplyr::group_by(model) |>
+    dplyr::summarize(
+      n_forecast_dates = dplyr::n(),
+      prop_present = n_forecast_dates / n_unique_forecasts
```
dylanhmorris:
```suggestion
      prop_present = n_forecast_dates / !!n_unique_forecasts
```

wweval/R/score.R
```diff
@@ -186,6 +186,109 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    )
+
+  models <- n_forecasts_per_model |>
+    dplyr::filter(prop_present > prop_dates_for_incl_hub) |>
```
dylanhmorris:
```suggestion
    dplyr::filter(prop_present > !!prop_dates_for_incl_hub) |>
```

wweval/R/score.R
dylanhmorris: Proportion of locations not implemented?

dylanhmorris (changes requested):
A few things.

Biggest question: how do you want to do proportion of locations present (which is not currently implemented, unless I'm missing something).

I think my instinct would be "at least n weeks of forecasts of at least y locations", i.o.w. a model only gets credit for a given week of forecasts (for `prop_present`) if it had at least  `prop_locs_for_incl_hub` of locations represented in its forecasts for that week.

_targets_eval_postprocessing.R
kaitejohnson: Yeah I think we can set this up to be specified in the config. Here I want to generate a figure to see all the models and then choose which ones we want to visualize so I will update this at some point and move to config after looking at results. Think we need to justify the choice of the one or two non ensemble models (i.e. best performing)

wweval/R/score.R
kaitejohnson: My bad I thought I moved it over from the scratch file

_targets_eval_postprocessing.R
dylanhmorris: Good point, but I meant more that this could be all of `covidhub_models_to_score` in the first instance

src/write_eval_config.R
dylanhmorris: Again, suggest being a bit more permissive here.

wweval/R/score.R
```diff
@@ -186,6 +186,116 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    types = "quantile",
+    timezeros = forecast_dates
+  )
+
+  n_unique_forecasts <- forecast_data |>
+    dplyr::distinct(timezero) |>
+    dplyr::pull() |>
+    length()
+
+  n_forecasts_per_model <- forecast_data |>
```
dylanhmorris:
```suggestion
  forecasts_present_per_model <- forecast_data |>
```

wweval/R/score.R
```diff
@@ -186,6 +186,116 @@ make_baseline_score_table <- function(all_ww_scores,
...
+    ) |>
+    # Exclude any forecast dates/models with too few locations submitted
+    dplyr::filter(prop_locs >= !!prop_locs_for_incl_hub) |>
+    dplyr::group_by(model) |>
+    dplyr::summarize(
+      n_forecast_dates = dplyr::n(),
+      prop_present = n_forecast_dates / !!n_unique_forecasts
+    )
+
+  models <- n_forecasts_per_model |>
```
dylanhmorris:
```suggestion
  models <- forecasts_present_per_model |>
```

dylanhmorris (dismissed):
LGTM. Optional minor wording and threshold suggestions

kaitejohnson:
Would you suggest 18/22 and 50/52? Fine with that, I don't think its likely to change the models chosen!

dylanhmorris:
> Would you suggest 18/22 and 50/52? Fine with that, I don't think its likely to change the models chosen!

I would suggest something more like 40/52. 18/22 seems good.

_targets_eval_postprocessing.R
kaitejohnson: Yeah, good point. We should probably include all of them now including the many ensembles, and then we can make a decision about which ensemble to include in the visualization. Will add them and add plotting colors for them. 

Another related question will be whether we want to include all 4 ensemble models in the scoring ranking for that component of figure 5.

wweval/R/score.R
```diff
@@ -190,6 +190,116 @@ make_baseline_score_table <- function(all_ww_scores,
...
+  zoltr::zoltar_authenticate(
+    zoltar_connection, get_secret("Z_USERNAME"),
+    get_secret("Z_PASSWORD")
+  )
+
+  # list of project on zoltar
+  the_projects <- zoltr::projects(zoltar_connection)
+
+  # Grabbing a specific project
+  project_url <- the_projects[the_projects$name == project_name, "url"]
```
kaitejohnson, line 255: Using `!!` caused this error: `Error in !project_name : invalid argument type`. I think this is ok as is

dylanhmorris: That's because this is a base filter rather than a `dplyr::filter()`, so it doesn't use tidyeval. Yes, it's fine.

kaitejohnson:
A little scope creep to get everything working needed to do the filtering to log scores. I am going to work off of this branch to remove the outlier wastewater data

wweval/R/ms_fig5.R
```diff
@@ -261,14 +261,14 @@ make_fig5_heatmap_relative_wis <- function(scores,
...
-    ) # exclued the US bc not available for
+    ) # exclude the US bc not available for
   # retro model
 
 
   p <- ggplot(relative_scores) +
     geom_tile(aes(x = model, y = short_name, fill = relative_interval_score)) +
     scale_fill_gradient2(
-      high = "red", mid = "white", low = "blue", transform = "log2",
+      high = "red", mid = "white", low = "blue",
```
dylanhmorris: why was the transform removed?

_targets_eval_postprocessing.R
```diff
@@ -1095,12 +1097,19 @@ hub_targets <- list(
   # Write a function that will get hub scores + all the metadata
   # horizon by week, location, forecast_date + eval data alongside it
   # for the models specified in the eval config
+  tar_target(
+    name = boolean_to_pull_locally,
+    command = any(is.null(c(
+      dim(metadata_hub_submissions),
+      dim(metadata_hosp_hub_submissions)
+    )))
```
dylanhmorris: Why does this need to be a separate target?

dylanhmorris (commented):
A couple questions @kaitejohnson

wweval/R/ms_fig5.R
kaitejohnson: This must be an issue with my local vs VAP installed version of `ggplot2` -- I can fix this to make the versions match/update on the VAP. 

In my VAP installation of ggplot2, `transform` was not an argulemt to `scale_fill_gradient2()` so this was throwing an error. Should've looked into it more but just excluded to generate the figure

_targets_eval_postprocessing.R
kaitejohnson: Ah yes, it wasn't originally but it needs to be because of the dependency graph -- the retrospective hub submission files need to be generated locally in order to be passed to `score_hub_submissions()` which looks for a file path with the Hub naming convention (either locally or from github). If you don't set this as a target and pass it to that function, it will not know that this needs to be run first (and the subsequent targets will fail bc they are missing the needed files)

wweval/R/ms_fig5.R
dylanhmorris: You probably need to update ggplot2. It was changed from `trans` to `transform` in later versions https://ggplot2.tidyverse.org/reference/scale_gradient.html

_targets_eval_postprocessing.R
dylanhmorris: Discussed a different sol'n face to face

wweval/R/utils.R
```diff
@@ -26,3 +26,15 @@ check_package_is_installed <- function(pkg_name) {
     )
   }
 }
+
+#' Helper function to pass along dependencies
```
dylanhmorris:
```suggestion
#' Helper function to note targets dependencies not explicitly 
#' noted in the function call
```

wweval/R/utils.R
```diff
@@ -26,3 +26,15 @@ check_package_is_installed <- function(pkg_name) {
...
+#'
+#' @param x any dependecy
```
dylanhmorris:
```suggestion
#' @param output to be passed along (typically the output of a function call
```

wweval/R/utils.R
```diff
@@ -26,3 +26,15 @@ check_package_is_installed <- function(pkg_name) {
...
+#' @param ... additional args
```
dylanhmorris:
```suggestion
#' @param ... additional arguments, which can be used to indicate additional upstream targets 
#' to treat as dependencies when this function is used in a target `command`
```

wweval/R/utils.R
```diff
@@ -26,3 +26,15 @@ check_package_is_installed <- function(pkg_name) {
...
+#'
+#' @return NULL
```
dylanhmorris:
```suggestion
#' @return the first argument
```

_targets_eval_postprocessing.R
```diff
@@ -1116,8 +1109,11 @@ hub_targets <- list(
         ),
         to = lubridate::ymd(max(eval_config$forecast_date_hosp)),
         by = "week"
-      )
-    )
+      ) # Ensure that local retrospective hub submission files have been made
+    ) |> with_dependencies(c(
+      metadata_hub_submissions,
+      metadata_hosp_hub_submissions
+    ))
```
dylanhmorris: Don't need to concatenate them since `with_dependencies` takes arbitrarily many additional arguments
```suggestion
    ) |> with_dependencies(
      metadata_hub_submissions,
      metadata_hosp_hub_submissions)
```

dylanhmorris (commented):
A few tweaks to clarify docs/intended usage

dylanhmorris (approved):
LGTM

PR #123, opened 2024-07-30
Move ordering functions to separate file

### LINKED ISSUE #121: Move ordering functions to their own file and checks for column existence
Comment here, want them to be in a separate file
https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/112#discussion_r1696124673

kaitejohnson:
This PR closes #121. It moves the ordering functions to a separate file and adds an error message if the columns in the dataframe don't match the expected values that are defined in the ordering.


dylanhmorris (approved):
LGTM, but note that this doesn't close #121 because we still need to column existence checks. Can merge this and then do that in a separate PR @kaitejohnson

PR #124, opened 2024-07-30
120 refactor to generate flags in postprocess

### LINKED ISSUE #120: Streamline azure postprocess and combination of ww data in `_targets_eval_postprocessing.R`
## Goal
Currently, we are using the post-processed outputs (ww quantiles joined to input data) to identify quality issues (high latency, extremely high noise, basically no sites, etc) where we would want to exclude that location-forecast date from the head-to-head comparison. This is not great, bc theres tons of wastewater data and its unnecessarily slow, and these flags are only a function of the input data

## Requirements
- [ ] modify `command_line_post_process()` and associated functions to compute an individual state-forecast-date's ww flags to determine if the wastewater data is sufficient
- [ ] modify `command_line_post_process()` and associated functions to save the input data for reproducibilty (both ww and hosp)
- [ ] modify `_targets_eval_postprocessing.R()` to pull in all of the `ww_data_flags` from across the nested file structure and get the table of sufficient ww data. Use this instead of the current workflow, which relies on the full set of quantiles and is therefore very slow
- [ ] only pull in the ww quantiles needed for plotting Fig 2, we don't need to hold all of them in memory

kaitejohnson:
This PR closes #120. It does the following:
- creates a new function that generates a 5 row table of wastewater data flags for each individual model run `get_ww_data_flags()`
- adds this to the postprocessing pipeline, and saves it in the nested folder structure alongside the input wastewater data and input hospital admissions data (even though these data are joined to the quantiled outputs)
- adds to the `_targets_eval_postprocessing.R` file to create a combined table of all the wastewater data flags (analogous to `_tar_combine()`)
- this combined wastewater data flags table then gets passed to `get_table_sufficient_ww()` which was rewritten to summarize this table, returning one boolean on whether the wastewater data is sufficient for each location-forecast_date combination
- removes the loading in of all of the wastewater quantiles, since this was very slow. Instead, we pull in only the quantiles that are needed for generating Figure 2. 
- removes the hacky things I added to avoid regenerating wastewater quantiles

@dylanhmorris I still need to test this with an Azure run today, but I think this is generally what we want.


kaitejohnson:
@dylanhmorris This is the issue I run into when building the container. https://github.com/cdcent/cfa-forecast-renewal-ww/issues/763

I am going to switch back to prod to make sure the issue isn't with this codebase, but is in fact just something with our `setup_container.R`

kaitejohnson:
Update I am testing this out now on a few states/forecast dates. Will ensure that it generates the outputs we expect, then will run on all locations. This should make the `_targets_eval_postprocessing.R` faster to run when outputs change bc it won't require that we load in all the wastewater data and the quantiles to get the dates/locations to exclude

kaitejohnson:
Confirmed that this generates a dataframe of flags within the nested folder structure for each individual run (just ran on an example few states and dates for now).

src/setup_eval.R
```diff
@@ -14,21 +14,20 @@ write_eval_config(
...
-      )
-    ),
+  forecast_dates = c("2024-02-05"),
+  as.character(
+    seq(
+      from = lubridate::ymd("2023-10-16"),
+      to = lubridate::ymd("2024-03-11"),
+      by = "week"
+    )
+  ),
```
dylanhmorris: Typo 

```suggestion
  forecast_dates =  as.character(
    seq(
      from = lubridate::ymd("2023-10-16"),
      to = lubridate::ymd("2024-03-11"),
      by = "week"
    )
  ),
```

wweval/R/get_table_sufficient_ww.R
```diff
@@ -27,29 +60,28 @@
...
       n_dps = dplyr::n(),
       prop_below_lod = sum(below_LOD == 1) / dplyr::n(),
-      sd = sd(calib_data),
-      mean_log_ww = mean(log(calib_data))
+      sd = sd(ww),
+      mean_log_ww = mean(log(ww))
     ) |>
     dplyr::mutate(
+      forecast_date = lubridate::ymd(!!forecast_date),
       flag_delay = as.integer(forecast_date - last_date) > delay_thres,
```
```suggestion
      flag_delay = as.integer(forecast_date - last_date) > !!delay_thres,
```

wweval/R/get_table_sufficient_ww.R
```diff
@@ -27,29 +60,28 @@
...
       flag_n_dps = n_dps < n_dps_thres,
```
```suggestion
      flag_n_dps = n_dps < !!n_dps_thres,
```

wweval/R/get_table_sufficient_ww.R
```diff
@@ -27,29 +60,28 @@
...
       flag_lod = prop_below_lod > prop_below_lod_thres,
```
```suggestion
      flag_lod = prop_below_lod > !!prop_below_lod_thres,
```

dylanhmorris: And also the `sd_thres` and `mean_log_ww_value_thres` below.

dylanhmorris (changes requested):
A couple minor but key fixes, then good to go @kaitejohnson

wweval/R/get_table_sufficient_ww.R
kaitejohnson: Good catch

dylanhmorris (approved):
LGTM!

PR #127, opened 2024-07-30
Fix epidemic phase analysis

### LINKED ISSUE #125: Fix the epidemic phase analysis to be based on the phase the forecast date is in
## Problem
Currently assigning the epidemic phase to individual target end dates. What we actually want to do is assign them to the forecast date, and then evaluate all the horizons in that forecast date as in an epidemic phase. I had misinterpreted how to implement this.

kaitejohnson:
This PR closes #125. Instead of joining the epidemic phases by date, we instead left join to forecast date and location -- so any scores from a single forecast date and location are assigned a single epidemic phase corresponding to the phase assigned using the retrospective R(t) (so with future knowledge). 

Revised figures [here](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/759#issuecomment-2258868022)


wweval/R/ms_fig4.R
```diff
@@ -484,13 +485,36 @@ make_fig4_rel_crps_by_phase <- function(scores) {
     geom_hline(aes(yintercept = 1), linetype = "dashed") +
     xlab("Epidemic phase") +
     ylab("Relative CRPS") +
-    scale_y_continuous(trans = "log10") +
+    scale_y_continuous(trans = "log10", limits = c(0.5, 2)) +
     get_plot_theme(
       x_axis_title_size = 8,
       y_axis_title_size = 8
     ) +
     scale_fill_brewer(palette = "Set3")
```
dylanhmorris: not for this PR, but worth creating a set palette for the phases too.

wweval/R/ms_fig4.R
```diff
@@ -484,13 +485,36 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
+      position = position_dodge(width = 0.75),
+    ) +
+    xlab("Epidemic phase") +
+    ylab("CRPS") +
+    scale_y_continuous(limits = c(0.0, 1)) +
+    get_plot_theme(
+      x_axis_title_size = 8,
+      y_axis_title_size = 8
+    ) +
+    scale_fill_manual(values = colors$model_colors)
```
dylanhmorris: If you want to hold onto this code within this file, it should be moved into a separate function (which just never gets called).

Related: I'd consider modularizing this function into a `compute_relative_crps` function and a `plot_relative_crps` function

dylanhmorris: First point should be addressed in this PR, second doesn't have to be.

_targets_eval_postprocessing.R
```diff
@@ -435,7 +435,7 @@ head_to_head_targets <- list(
         epidemic_phases,
         by = c(
           "location" = "state_abbr",
-          "date" = "reference_date"
+          "forecast_date" = "reference_date"
```
dylanhmorris: question: why is the relevant date column of the `epidemic_phases` dataframe called `reference_date`? Is that what it's called in the source file or do we rename it that somewhere?

dylanhmorris (changes requested):
Looks largely good. One key change and a couple suggestions/questions

wweval/R/ms_fig4.R
kaitejohnson: Yeah thats a good point. Will make a second PR for the `compute_relative_crps` fxns

_targets_eval_postprocessing.R
kaitejohnson: Ah yeah this is from the retrospective R(t) and I just kept NNH's nomenclature (they use reference date and report date). I can change it when `epidemic_phases` gets made though

wweval/R/ms_fig4.R
kaitejohnson: https://github.com/CDCgov/wastewater-informed-covid-forecasting/issues/131

_targets_eval_postprocessing.R
kaitejohnson: Fixed this

wweval/R/ms_fig4.R
dylanhmorris: #134

wweval/R/ms_fig4.R
```diff
@@ -484,14 +486,78 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
       x_axis_title_size = 8,
       y_axis_title_size = 8
     ) +
-    scale_fill_brewer(palette = "Set3")
+    scale_fill_manual(values = colors$phase_colors)
+
+  return(p)
+}
+#' Make figure that plots distribution of crps scores stratified by
+#' epidemic phase
```
dylanhmorris, line 499:
```suggestion
#' epidemic phase but aggregated across horizons
```

wweval/R/ms_fig4.R
```diff
@@ -484,14 +486,78 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
+#' CRPS scores, colored by model, stratified by epidemic phase, across
+#' locations and forecast dates
+#' @export
+make_sfig_crps_by_phase <- function(scores) {
+  scores_w_fig_order <- scores |>
+    order_phases()
+
+  # Quick warning if there are NAs in epidemic phases
+  missing_phases <- scores |>
+    dplyr::filter(is.na(phase))
```
dylanhmorris, line 515: This is supposed to reproduce the older plot that was aggregated across horizons within a location, yes? Perhaps I'm missing something, but I don't think it will. The code looks equivalent to the disaggregated main figure above.

dylanhmorris (changes requested):

wweval/R/ms_fig4.R
kaitejohnson: I accidentally copied over the computing of the relative crps but it isn't being used. The difference here is that we are making the stathalfeye plot colored by model and with the y axis value by crps, instead of relative crps

wweval/R/ms_fig4.R
kaitejohnson: wait, these are not aggregated across horizons. They are just plots of the distribution of scores for each loc-forecast-date-day combination, without any summarizing across the horizon

wweval/R/ms_fig4.R
dylanhmorris: Got it, thanks!

wweval/R/ms_fig4.R
```diff
@@ -484,14 +486,60 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
     get_plot_theme(
       x_axis_title_size = 8,
       y_axis_title_size = 8
     ) +
-    scale_fill_brewer(palette = "Set3")
+    scale_fill_manual(values = colors$phase_colors)
+
+  return(p)
+}
+#' Make figure that plots distribution of crps scores stratified by
```
dylanhmorris:
```suggestion
#' Make figure that plots distribution of absolute crps stratified by
```

wweval/R/ms_fig4.R
```diff
@@ -484,14 +486,60 @@ make_fig4_rel_crps_by_phase <- function(scores) {
...
+#' epidemic phase
+#'
+#' @param scores A tibble of scores by location, forecast date, date and model,
+#' containing the outputs of `scoringutils::score()` on samples plus metadata
+#' transformed into a tibble.
+#'
+#' @return A ggplot object containing plots of the distribution of
+#' CRPS scores, colored by model, stratified by epidemic phase, across
```
dylanhmorris:
```suggestion
#' continuous ranked probability scores, colored by model, stratified by epidemic phase, across
```

dylanhmorris (approved):
LGTM. A couple wording tweaks you can take or leave.

PR #132, opened 2024-07-31
Create a separate function for computing the relative CRPS

### LINKED ISSUE #131: Make a `compute_relative_crps()` function
## Goal
We do this in a lot of figures and its repetitive, make it a function 

https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/127#discussion_r1697643728

kaitejohnson:
This PR closes #131. This function is called in most of the figure 4 plots to compute relative CRPS at each location-forecast_date -date. Added some checks to ensure the necessary column names and model types are in the scores table that gets passed to it.


seabbs (commented):
why not use the functionality in `scoringutils` to do this (in summarise_score?)

kaitejohnson:
> why not use the functionality in `scoringutils` to do this (in summarise_score?)

Bc I forgot it existed, let me take a look!

kaitejohnson:
We have a `n` row table of scores for each location and forecast date for two models, and we want `n/2` relative scores for just the wastewater model compared to the admissions model. 

When I try `summarize_scores(relative_skill = TRUE, baseline = "hosp")` (and then presumably we would just filter out the `model=='hosp'` since these will all be 1, I get this error:
```
Warning message:
In check_summary_params(scores = scores, by = by, relative_skill = relative_skill,  :
  you need more than one model non-baseline model to make model comparisons. Relative skill will not be computed
  ```

I think it might be just as easy in this case to pivot the data from long to wide and then divide the ww score by the hosp score? @seabbs

**Comment by dylanhmorris** [comment]
> When I try summarize_scores(relative_skill = TRUE, baseline = "hosp") (and then presumably we would just filter out the model=='hosp' since these will all be 1, I get this error:

What `scores` dataframe are you using as the input to that call?

**Comment by kaitejohnson** [comment]
> > When I try summarize_scores(relative_skill = TRUE, baseline = "hosp") (and then presumably we would just filter out the model=='hosp' since these will all be 1, I get this error:
> 
> What `scores` dataframe are you using as the input to that call?

This is the `scores_filtered` target in our pipeline, which is the long tibble containing all scores from all forecast dates, dates, locations, and the two models, excluding models with non-convergence or missing wastewater data

File: wweval/R/compute_relative_crps.R [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+      c("Model names are not `ww` and `hosp` as function expects")
+    )
+  }
+
+
+  rel_crps <- scores |>
+    tidyr::pivot_wider(
+      names_from = model,
+      values_from = crps,
+      id_cols = !!id_cols
```
dylanhmorris:
```suggestion
      id_cols = {{ id_cols }}
```
If you do it this way you can use either raw words or strings, i.e.

`id_cols = c(location, forecast_date, horizon)`
or 
`id_cols = c("location", "forecast_date", "horizon")`

dylanhmorris (approved):
One suggestion you can take or leave. Can defer doing things via `scoringutils` to a separate PR.

wweval/R/compute_relative_crps.R
kaitejohnson: Oh that is nice

PR #136, opened 2024-08-16
fix back calculation in R(t)

kaitejohnson:
Realized that the forward model assumes no infectiousness on day 0 of infection, and this is not the case if you just back-calculate R(t) by convolving incident infections with the generation interval specified, bc this generation interval is starting at day 1. 

To back-calculate properly, we need to add 0 density on day 0 to the generation interval.

This explains why our R(t)s have looked jagged.


cfaforecastrenewalww/inst/stan/renewal_ww_hosp_site_level_inf_dynamics.stan
```diff
@@ -365,6 +365,7 @@ generated quantities {
 
   // Deterministic calculation of state level R(t) from incident infections
   // and the generation interval
-  rt = (state_inf_per_capita ./ convolve_dot_product(state_inf_per_capita, gt_rev_pmf, uot + ot + ht))[uot+1: uot + ot + ht];
+  rt = (state_inf_per_capita ./ convolve_dot_product(state_inf_per_capita,
+    reverse(append_row(0,generation_interval)), uot + ot + ht))[uot+1: uot + ot + ht];
```
dylanhmorris, line 369: Do you definitely want to do this with `convolve_dot_product` and not `update_infectiousness`?

dylanhmorris (commented):
One question @kaitejohnson, cc @seabbs

cfaforecastrenewalww/inst/stan/renewal_ww_hosp_site_level_inf_dynamics.stan
kaitejohnson: I think so, `update_infectiousness` is needed because it is tricky to do the convolution without an incident case curve, requires the initialization piece (from the exponential growth initial burn in period). 

But because we have the full incident infection time series, the calculation of R(t) should be  straightforward. The convolution works just like the other convolutions where we have a full time series and we convolve with a delay distribution. The kicker is just that those other delay distributions do have a value at index 0 (even if it is 0), whereas the GI pmf is assumed to start at day 1 not day 0.

kaitejohnson:
@dylanhmorris Is this good to merge?

dylanhmorris (approved):
LGTM

PR #138, opened 2024-08-26
Hot fix flags and targets

kaitejohnson:
@dylanhmorris I realized a few silly errors in some of the changes from #120. I forgot to add the location to the `ww_data_flags` table and combine outputs is set up to do just a straight rowbind, so it doesn't pass along location metadata by default. 

Also realized a small bug in the name of the file I was looking for when grabbing the wastewater quantiles.

### DIFF
--- a/wweval/R/get_table_sufficient_ww.R
+++ b/wweval/R/get_table_sufficient_ww.R
@@ -73,7 +73,7 @@ get_ww_data_flags <- function(input_ww_data,
                               sd_thres = 0.1,
                               mean_log_ww_value_thres = -4) {
   this_location <- input_ww_data |>
-    dplyr::select(location) |>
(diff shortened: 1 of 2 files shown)


wweval/R/get_table_sufficient_ww.R
```diff
@@ -81,6 +85,7 @@ get_ww_data_flags <- function(input_ww_data,
       mean_log_ww = mean(log(ww))
     ) |>
     dplyr::mutate(
+      location = this_location,
```
```suggestion
      location = !!this_location,
```

dylanhmorris (commented):
One key change

kaitejohnson:
Whoops I also used `select` instead of `distinct`

dylanhmorris (approved):
LGTM

PR #139, opened 2024-08-27
Scoring was failing because of required column names

kaitejohnson:
I am really not sure how this worked previously but we have done some refactoring so maybe there was a step that previously renamed value to prediction and eval data to true value....


kaitejohnson:
Confirmed this worked with a small azure run

dylanhmorris (approved):
LGTM provided it worked in a test run. Will let you merge @kaitejohnson

PR #140, opened 2024-08-28
Filter to only log scores

kaitejohnson:
We previously added functionality to pass around natural scale and log scale scores-- had forgot to filter for just log scale scores.

We could also set this up as a config variable, but I think this is okay for now


kaitejohnson:
Going to close this and incorporate in #119 so I can test everything

PR #141, opened 2024-08-28
Fix pre-commit

### LINKED ISSUE #144: Pre-commit doesn't appear to be running
Looks like we messed something up with the caching when moving the repo from cdcent to cdcgov. We solved this in the `wwinference` repo, and I think we just need to set up the cache properly in the `.precommit-config.yaml`

image

@zsusswein Had a similar[ issue-](https://github.com/CDCgov/cfa-gam-rt/issues/24)- might be worth making an issue to create a template repository for cdcgov if this is a common gotcha

kaitejohnson:
This PR closes #144

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,10 +28,10 @@ repos:
     -   id: no-print-statement
     -   id: no-debug-statement
     -   id: deps-in-desc
-        args: [--root= /wweval, --allow_private_imports]
+        args: [--root=wweval, --allow_private_imports]
         exclude: data-raw
     -   id: deps-in-desc
-        args: [--root= /cfaforecastrenewalww]
+        args: [--root=cfaforecastrenewalww]
         exclude: data-raw
 
 ########
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,10 @@ repos:
     -   id: no-print-statement
     -   id: no-debug-statement
     -   id: deps-in-desc
+        args: [--root=wweval] 
+        exclude: data-raw
+    -   id: deps-in-desc
+        args: [--root=cfaforecastrenewalww] 
         exclude: data-raw
 
 ########
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,17 @@ repos:
     rev: v0.4.2
     hooks:
     -   id: style-files
+        args: [--style_pkg=styler, --style_fun=tidyverse_style,
+               --cache-root=styler-perm]
+    -   id: use-tidy-description
     -   id: lintr
+    -   id: readme-rmd-rendered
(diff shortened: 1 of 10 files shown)


kaitejohnson:
Stuck here: 
```
Error in find_package_root(dir) : 
  Could not find R package in `.` or its parent directories.
```

.pre-commit-config.yaml
```diff
@@ -18,7 +18,17 @@ repos:
...
+               --cache-root=styler-perm]
+    -   id: use-tidy-description
     -   id: lintr
+    -   id: readme-rmd-rendered
+    -   id: parsable-R
+    -   id: no-browser-statement
+    -   id: no-print-statement
+    -   id: no-debug-statement
+    -   id: deps-in-desc
+        exclude: data-raw
```
kaitejohnson: Perhaps we need to exclude additional folders in this repo?

kaitejohnson: > Perhaps we need to exclude additional folders in this repo?

kaitejohnson:
This is basically flagging every package even if it is in the DESCRIPTION. 

Seems like the `root =` thing just isn't working because it seems to be scanning the whole repo (e.g. its flagging things in scratch)

kaitejohnson:
Not an ideal solution but wondering if for now we should just get rid of `deps-in-desc` because I think it is covered in `R-cmd-check`

zsusswein:
> Stuck here:
> 
> ```
> Error in find_package_root(dir) : 
>   Could not find R package in `.` or its parent directories.
> ```

This is super weird because I literally copied and pasted your `.pre-commit.config.yaml` and `.github/workflows/pre-commit.yaml` and it's working great

kaitejohnson:
> > Stuck here:
> > ```
> > Error in find_package_root(dir) : 
> >   Could not find R package in `.` or its parent directories.
> > ```
> 
> This is super weird because I literally copied and pasted your `.pre-commit.config.yaml` and `.github/workflows/pre-commit.yaml` and it's working great

I think the issue is that repo != package, but I am not sure why specifying the `root =<packagedir> isn't working, and have tried a few different things. 

@dylanhmorris My suggested way forward here is to remove the hook for `deps-in-desc` from the `.pre-commit-config.yaml`, which of course means they won't run locally but since they are in `check-ww` and `check-eval` `R CMD Check`, they will be caught in CI. I will make a separate issue to figure out how to add them back into pre-commit.

dylanhmorris:
> @dylanhmorris My suggested way forward here is to remove the hook for deps-in-desc from the .pre-commit-config.yaml, which of course means they won't run locally but since they are in check-ww and check-eval R CMD Check, they will be caught in CI. I will make a separate issue to figure out how to add them back into pre-commit.

Seems sensible to me.

dylanhmorris (approved):
LGTM

PR #143, opened 2024-08-28
Manual ww data exclusion

kaitejohnson:
This PR adds functionality to manually exclude specific wastewater forecast dates and locations from the analysis. I have applied this to MN in January/early February which is when we saw the jump in values in the time stamped datasets, likely due to a reporting error. 

I also caught some issues with how we were doing the Hub analysis. These were:
- needed to filter to include only the forecast dates and locations that we "would have submitted" based on presence of sufficient wastewater data and model convergence (same as in the head to head analysis, as described in manuscript)
- needed to filter scores when we used the `scoringutils` function to log transform observations and predictions, since this returns scores both in natural and log scale


dylanhmorris (dismissed):
LGTM. My only question is whether we want to name the table `ww_data_exclusions` versus something else. To my ear, the current name sounds more like a set of datapoints to exclude than a set of forecast dates to exclude overall due to data problems. Perhaps `ww_forecast_dates_to_exclude`?

Can be a separate PR.

kaitejohnson:
I can change it here, perhaps just `ww_forecast_date_locs_to_exclude`  since we don't want to imply we are excluding the entire forecast date, just the entire forecast date-location combination? Agree it is too similar to the hospital admissions one which is specific data points

dylanhmorris:
>  `ww_forecast_date_locs_to_exclude`

Sounds good.

_targets_eval_postprocessing.R
```diff
@@ -416,6 +436,14 @@ head_to_head_targets <- list(
       dplyr::left_join(table_of_loc_dates_w_ww,
         by = c("location", "forecast_date")
       ) |>
+      dplyr::anti_join(
+        table_of_ww_exclusions |>
```
dylanhmorris:
```suggestion
        ww_forecast_date_locs_to_excl |>
```

_targets_eval_postprocessing.R
```diff
@@ -383,6 +395,14 @@ head_to_head_targets <- list(
...
```
dylanhmorris: These need to be changed too, no?
```suggestion
        ww_forecast_date_locs_to_excl |>
```

dylanhmorris (changes requested):

_targets_eval_postprocessing.R
kaitejohnson: Oops my bad good catch

dylanhmorris (approved):
LGTM

PR #148, opened 2024-08-30
Fix hierarchical stdev bug

### LINKED ISSUE #145: Duplicate error in hierarchical sd
## Problem
Currently, the observation model on wastewater concentrations is:

$$\log[c_{ijt}] \sim \mathrm{Normal}(\log[M_{ij} C_i(t)], \sigma_{cij})$$

Where $C_i(t)$ is the expected concentration in the subpopulaiton i,  $M_{ij}$ is a site-labl level multiplier and $\sigma_{cij}$ is the standard deviation in the log concentrations in a site-lab. Both $M_{ij}$ and $\sigma_{cij}$ are estimated hierarchically as (originally):

$$\log(M_{ij}) \sim \mathrm{Normal}(0, \sigma_m)$$

$$\sigma_{cij} \sim \mathrm{Normal}(\mu_{\sigma},\sigma_{\sigma_c})$$

With priors originally of:

$$\sigma_m \sim N(0,0.25) $$

$$ \sigma_{\sigma_c} \sim N(0,1) $$

$$\mu_{\sigma} \sim N(0,1) $$

The problem with this implementation is that it permits sampling negative $\sigma_{cij}$ s. As written, the prior assumes that ~95% of observations will be between -4 and 4 x the logmean concentration, which is relatively diffuse. 

## Solution
 My understanding is that this is what was implemented in #117 , and this issue is to fix the model definition and confirm understanding. 

We'd have the same observation model as written: 

$$\log[c_{ijt}] \sim \mathrm{Normal}(\log[M_{ij} C_i(t)], \sigma_{cij})$$

where:

$$\log(\sigma_{cij}) \sim \mathrm{Normal}(\log(mode_{\sigma}), \sigma_{ c, \log})$$ 

Where the non-centered parameterization of $\sigma_{cij}$ can be written as:

$$ \sigma_{\sigma_c}  = exp(\log(mode_{\sigma}) +  \sigma_{ c, \log}\mathrm{Normal}(0,1))$$

And the priors are:

$$\log(mode_{\sigma}) \sim \mathrm{Normal}(1,1)$$

$$  \sigma_{ c, \log} \sim \mathrm{Normal}(0, log(2))$$

Which with this prior would mean that ~95% of observations will be between 1/4 and 4x the logmean concentration

Copied from: https://github.com/CDCgov/ww-inference-model/issues/119

kaitejohnson:
This duplicates the hierarchical sd bug fix performed here https://github.com/CDCgov/ww-inference-model/pull/117 and closes #145

Note this does not fix the model definition, I will do so in a separate PR once this is closed https://github.com/CDCgov/ww-inference-model/pull/120


input/params.toml
```diff
@@ -59,10 +59,10 @@ log_g_prior_sd = 4.60517 # 2 * log(10)
...
 sigma_ww_prior_sd = 0.5
 
-sigma_ww_site_prior_mean_mean = 1
-sigma_ww_site_prior_mean_sd = 1
-sigma_ww_site_prior_sd_mean = 0
-sigma_ww_site_prior_sd_sd = 1
+mode_sigma_ww_site_prior_mode = 1
+mode_sigma_ww_site_prior_sd = 1
+sd_log_sigma_ww_site_prior_mode = 0
+sd_log_sigma_ww_site_prior_sd = 0.693 # log(2)
```
seabbs, line 65: why not just write log(2)? the compute is tiny

kaitejohnson: Can tomls do compute?

dylanhmorris: Not internally. For that reason, I think it's fine as is.

dylanhmorris (approved):
LGTM

PR #149, opened 2024-08-30
forgot to update one parameter name

kaitejohnson:
Testing with azure runs found another place where we needed to use the new way of naming the mode stdev parameter


wweval/R/format_data_for_stan.R
```diff
@@ -308,7 +308,7 @@ get_stan_data_list <- function(model_type,
       wday_effect_prior_sd = wday_effect_prior_sd,
       initial_growth_prior_mean = initial_growth_prior_mean,
       initial_growth_prior_sd = initial_growth_prior_sd,
-      sigma_ww_prior_mean = sigma_ww_site_prior_mean_mean,
+      sigma_ww_prior_mean = mode_sigma_ww_site_prior_mode,
```
dylanhmorris, line 311: Shouldn't this be
```suggestion
      mode_sigma_ww_site_prior_mode = mode_sigma_ww_site_prior_mode,
```
?

kaitejohnson: No, so this is in the hospital admissions only model switch (which calls the `renewal_ww_hosp.stan` model, which as written, has a sigma_ww for the single wastewater observation model it was originally designed to fit). This wouldn't be used.

kaitejohnson: Confusing I know. And a reason to move to using the package eventually! But for now we have these two workflows

dylanhmorris (approved):
LGTM

PR #151, opened 2024-09-04
150: Make wastewater metadata and summary stats 

### LINKED ISSUE #150: Generate a table of wastewater metadata
## Goal
We want to have some high level metadata on the wastewater data. This includes presence/absence of wastewater data across forecast dates and jurisdictions, and when wastewater data is present, some metrics on the wastewater data in that state including:
- Number of sites 
- Population coverage 
- Average latency in reporting 
- Average reporting frequency 

Propose that we do this by adding to the pipeline the creation of a table that generates this metadata for every forecast date-location. If the forecast-date location is missing wastewater, most will be NAs. 

We can then use this to make some hypotheses about why/how forecast performance varies across jurisdictions (and potentially add an analysis of these metrics vs forecast performance more generally)

kaitejohnson:
Ignore the branch name, I was working off of the hot fix branch and forgot to create a new one. Since they were very small changes decided to just leave on this branch.

This closes #150 

What this does:
- writes a new function that loads in each `input_ww_data` (which is now saved as part of the model fitting process), and computes wastewater metadata metrics for that particular forecast date and location (e.g. population coverage, number of sites, number of labs, reporting latency, sampling frequency etc.). Binds a row iteratively to a table so that we can avoid pulling everything into memory and returns a metadata table of length (n_locations x n_forecast dates = 1144)
- writes a new function that combines this metadata at the forecast-date-location level with the other metadata on: manual exclusion of wastewater data, wastewater sufficient quality, and model convergence
- writes a new function that summarized this granular metadata table : overall, across locations by forecast date, and across forecast dates by location. This should be particularly useful for the discussion section where we are hypothesizing about variation in performance across locations and forecast dates -- we will be able to use this to get a high level sense of the wastewater data in each jurisdiction/at each time point. This also helps us summarize the number of forecasts with and without wastewater and for what reasons. See summary and discussion in [internal repo](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/768) where I've posted the results of the tables from the current analysis.
- two bug fixes to targets pipeline

Let me know if there is additional granularity/summary metrics to add.


wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,368 @@
...
+      scenario = "status_quo",
+      forecast_date = this_forecast_date,
+      model_type = "ww",
+      location = this_location,
+      output_type = glue::glue("input_hosp_data"),
+      file_extension = "tsv"
+    )
+
+    if (file.exists(fp)) {
+      tryCatch(
```
dylanhmorris: The long `tryCatch` here suggests to me this should be broken into a helper function that is called iteratively and fails gracefully (e.g. via wrapping in `purrr::safely`)

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,368 @@
...
+#' @param convergence_df table containing convergence flags for every forecast
+#' date location combination
+#' @param table_of_loc_dates_w_ww table containing wastewater metadata
+#' for every location-forecast date with wastewater
+#'
+#' @return a tibble with a number of additional columns indicating whether
+#' or not there were any flags for manual exclusions, convergence issues,
+#' or wastewater quality issues
+#' @export
+get_add_ww_metadata <- function(granular_ww_metadata,
```
dylanhmorris, line 251: This function name is a bit opaque, as is the argument name `granualar_ww_metadata`.

It seems like the key thing that's actually going on here is joining wastewater data quality metrics (which are based on input data) to run convergence flags (which are based on mcmc output). Is that correct?

The purpose of the joins with `ww_forecast_date_locs_to_excl` and `table_of_loc_dates_w_ww` is currently also a bit opaque in their purpose. Is the aim to add a "was this manually excluded" column and a "was there sufficient wastewater on this date" column, respectively?

dylanhmorris (changes requested):
Some questions/requested changes before I review in more depth

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: Yeah so `get_add_ww_metadata()` just joins the other wastewater exclusions to the one row per forecast-date-location table (`granular_ww_metadata`)

kaitejohnson: And yeah, its to be able to slice and dice across all of those reasons and get summaries

kaitejohnson: `get_granular_ww_metdata()` makes a large table with one row per location-foercast-date with a bunch of metadata for that particular wastewater data

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+#'
+#' @return ww_metadata: a tibble with a row for each forecast date location and
+#' columns that provide summaries of the wastewater data
+#' @export
+combine_and_summarize_ww_data <- function(forecast_dates,
+                                          locations,
+                                          eval_output_subdir) {
+  if (length(forecast_dates) != length(locations)) {
+    cli::cli_abort(
+      message = "Vector of forecast dates and locations must be equal in length"
```
dylanhmorris:
```suggestion
      message = c("Vector of forecast dates and vector of locations must be equal in length",
                             "Got {length(forecast_dates)} forecast dates and {length(locations)} locations")
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+    this_forecast_date <- forecast_dates[i]
+    this_location <- locations[i]
+
+    fp_ww <- get_filepath(
+      eval_output_subdir,
+      scenario = "status_quo",
+      forecast_date = this_forecast_date,
+      model_type = "ww",
+      location = this_location,
+      output_type = glue::glue("input_ww_data"),
```
dylanhmorris: why does this need to be `glue::glue`?

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      file_extension = "tsv"
+    )
+    fp_hosp <- get_filepath(
+      eval_output_subdir,
+      scenario = "status_quo",
+      forecast_date = this_forecast_date,
+      model_type = "ww",
+      location = this_location,
+      output_type = glue::glue("input_hosp_data"),
```
dylanhmorris: As above, why does this need to be `glue::glue`?

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      file_extension = "tsv"
+    )
+
+    if (file.exists(fp_ww)) {
+      safe_data_load <- purrr::safely(load_data_and_summarize)
```
dylanhmorris: This will recreate the safe version of the function at every step of the for-loop. I would create it outside this function (i.e. at the scope of the `combine_and_summarize_ww_data.R` file, and just not export it if you don't think it should be user-exposed.

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      out <- safe_data_load(fp_hosp, fp_ww, this_forecast_date, this_location)
+
+      # If it doesn't error, then get metadata
+      if (is.null(out$error)) {
+        this_ww_metadata <- out$result
+      } else { # if it errored but file exists, this means that ww data was empty
```
dylanhmorris: Not good to rely on an errors for _expected_ behavior. Why can the function `load_data_and_summarize` not have a way to identify valid-but-empty wastewater data and return 

```R
tibble::tibble(
          forecast_date = this_forecast_date,
          location = this_location,
          ww_data_present = 0,
          n_sites = NA,
          n_labs = NA,
          pop_coverage = NA,
          state_pop = state_pop,
          avg_latency = NA,
          avg_sampling_freq = NA,
          n_duplicate_obs = NA
)
```

when it does?

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      "files_missing", model_type
+    ))
+
+    write.csv(
+      flag_failed_output,
+      file.path(
+        eval_output_subdir, "files_missing", model_type,
+        glue::glue("ww_data_metadata.csv")
+      )
+    )
```
dylanhmorris: I thought we were using `readr` functions?

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      flag_failed_output <- rbind(flag_failed_output, this_failed_output)
+    } # end if else if filepath doesn't exist
+  } # end for loop
+
+  if (nrow(flag_failed_output) != 0) {
+    # Save the missing files in a new subfolder in the eval_output_subdir
+    cfaforecastrenewalww::create_dir(file.path(
+      eval_output_subdir,
+      "files_missing", model_type
+    ))
```
dylanhmorris, line 87: Do you definitely want to write the file only if there are failures, or do you want to write an empty file when there are none. Minor point either way.

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+}
+
+#' Get additional wastewater metadata
+#'
+#' @description
+#' This function takes the summary statistics from just the input wastewater
+#' data (granular_ww_metadata) and joins the downstream metadata about
+#' convergenece and manual exclusions and combines them all into one table
+#'
+#'
```
dylanhmorris:
```suggestion
```

dylanhmorris: Extra line

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+  granular_ww_metadata_used <- granular_ww_metadata |>
+    dplyr::mutate(forecast_date = lubridate::ymd(forecast_date)) |>
+    dplyr::left_join(
+      ww_forecast_date_locs_to_excl |>
+        mutate(
+          ww_exclude_manual = 1,
+          forecast_date = lubridate::ymd(forecast_date)
+        ),
+      by = c("location", "forecast_date")
+    ) |>
```
dylanhmorris: As written this will have `ww_exclude_manual = NA` whenever data was _not_ manually excluded. I think `ww_exclude_manual = FALSE` makes more sense in that case

```suggestion
    dplyr::left_join(
      ww_forecast_date_locs_to_excl |>
        mutate(
          ww_exclude_manual = TRUE,
          forecast_date = lubridate::ymd(.data$orecast_date)
        ),
      by = c("location", "forecast_date")
    ) |>
    dplyr::mutate(
       ww_exclude_manual = dplyr::replace_na(.data$ww_exclude_manual, FALSE)
    ) |>
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+#' @return a tibble with a number of additional columns indicating whether
+#' or not there were any flags for manual exclusions, convergence issues,
+#' or wastewater quality issues
+#' @export
+get_add_ww_metadata <- function(granular_ww_metadata,
+                                ww_forecast_date_locs_to_excl,
+                                convergence_df,
+                                table_of_loc_dates_w_ww) {
+  granular_ww_metadata_used <- granular_ww_metadata |>
+    dplyr::mutate(forecast_date = lubridate::ymd(forecast_date)) |>
```
dylanhmorris: Why are some of tables' date columns coerced with `lubridate::ymd()` but not others?

dylanhmorris:
```suggestion
    dplyr::mutate(forecast_date = lubridate::ymd(.data$forecast_date)) |>
```

dylanhmorris: Also, don't forget the `.data` pronoun

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+#' date and a summary by state
+#' @export
+get_summary_ww_table <- function(ww_metadata, hosp_quantiles_filtered) {
+  # First, get the true number of forecast-date locations with wastewater
+  # in the current analysis
+  n_w_ww_actual <- hosp_quantiles_filtered |>
+    dplyr::filter(
+      model_type == "ww"
+    ) |>
+    dplyr::distinct(forecast_date, location) |>
```
dylanhmorris:
```suggestion
    dplyr::distinct(.data$forecast_date, .data$location) |>
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+#'
+#' @return a list containing a summary overall table, a summary by forecast
+#' date and a summary by state
+#' @export
+get_summary_ww_table <- function(ww_metadata, hosp_quantiles_filtered) {
+  # First, get the true number of forecast-date locations with wastewater
+  # in the current analysis
+  n_w_ww_actual <- hosp_quantiles_filtered |>
+    dplyr::filter(
+      model_type == "ww"
```
dylanhmorris:
```suggestion
      .data$model_type == "ww"
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+
+  n_combos_w_ww_data <- ww_metadata |>
+    dplyr::summarise(
+      n_ww_present = sum(ww_data_present, na.rm = TRUE)
+    )
+
+  n_states_w_complete_ww_data <- ww_metadata |>
+    dplyr::group_by(location) |>
+    dplyr::summarize(
+      complete_ww = all(ww_data_present == 1)
```
dylanhmorris:
```suggestion
      complete_ww = all(.data$ww_data_present == 1)
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+    ) |>
+    ungroup() |>
+    dplyr::summarize(n_complete_ww = sum(complete_ww)) |>
+    dplyr::pull(n_complete_ww)
```
dylanhmorris:
```suggestion
    dplyr::pull(.data$n_complete_ww)
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,413 @@
...
+      n_ww_present = sum(ww_data_present, na.rm = TRUE)
+    )
+
+  n_states_w_complete_ww_data <- ww_metadata |>
+    dplyr::group_by(location) |>
+    dplyr::summarize(
+      complete_ww = all(ww_data_present == 1)
+    ) |>
+    ungroup() |>
+    dplyr::summarize(n_complete_ww = sum(complete_ww)) |>
```
dylanhmorris:
```suggestion
    dplyr::summarize(n_complete_ww = sum(.data$complete_ww)) |>
```

dylanhmorris (changes requested):
Major suggestion about not relying on errors. Also, check tidyeval conventions throughout. Noted some, but not all.

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: I can change this, I might have just forgot as we were definitely using both at one point

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: I think it's fine to just write if there are errors

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: Removed the safe altogether per our discussion. `load_and_summarize_data()` now returns a single row dataframe with the NAs for ww metrics if the wastewater data is an empty csv. 

This means that if there is another error, the pipeline will fail loudly (which is what we want here, bc it means the issue is something else).

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: edited

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+        prev_date = dplyr::lag(.data$date, 1),
+        diff_time = as.numeric(lubridate::days(
+          difftime(.data$date, prev_date)
+        ), "days")
+      ) |>
+      dplyr::ungroup() |>
+      dplyr::summarize(
+        mean_collection_freq = mean(.data$diff_time, na.rm = TRUE)
+      ) |>
+      dplyr::pull(.data$mean_collection_freq)
```
dylanhmorris: This isn't exactly the mean frequency. It's the mean gap between samples until the last. Simpler and closer to what we want to compute `n_days_with_samples / n_days_total`, no?

To see why these are different, consider an extreme case: a site with only two observed days in our 28 day period, day 1 and day 2. Its mean "frequency" on this metric is going to be 1, even though only one out of 14 days (and none of the final 26) were sampled. It will also give the result `NaN` when there's only one date, since the `prev_date` will be `NA`.

```R
single_date_df <- tibble::tibble(date="2024-01-01")

single_date_df |>
dplyr::mutate(
        prev_date = dplyr::lag(.data$date, 1),
        diff_time = as.numeric(lubridate::days(
          difftime(.data$date, prev_date)
        ), "days")
      ) |>
      dplyr::ungroup() |>
      dplyr::summarize(
        mean_collection_freq = mean(.data$diff_time, na.rm = TRUE)
      ) |>
      dplyr::pull(.data$mean_collection_freq)
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+    dplyr::pull(.data$n_hosp_flags)
+  n_combos_w_ww_conv_flags <- ww_metadata |>
+    dplyr::summarise(n_ww_flags = sum(.data$any_flags_ww, na.rm = TRUE)) |>
+    dplyr::pull(.data$n_ww_flags)
+
+  n_insuff_ww <- ww_metadata |>
+    dplyr::summarise(
+      n_ww_insuff = sum(.data$ww_sufficient == FALSE, na.rm = TRUE)
+    ) |>
+    dplyr::pull(n_ww_insuff)
```
dylanhmorris:
```suggestion
    dplyr::pull("n_ww_insuff")
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+  n_ww_excluded <- ww_metadata |>
+    dplyr::summarise(
+      n_ww_excl = sum(.data$ww_exclude_manual, na.rm = TRUE)
+    ) |>
+    dplyr::pull(.data$n_ww_excl)
+
+  n_w_ww_expected <- ww_metadata |>
+    dplyr::mutate(
+      ww_expected = dplyr::case_when(
+        ww_data_present == 1 & is.na(.data$ww_exclude_manual) &
```
dylanhmorris:
```suggestion
        ww_data_present == 1 & !(.data$ww_exclude_manual) &
```

dylanhmorris: This is now a boolean, no?

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+          !isTRUE(.data$any_flags_hosp) & !isTRUE(.data$any_flags_ww) &
+          isTRUE(.data$ww_sufficient) ~ TRUE,
```
dylanhmorris: In this situation, I think you probably want just plain `!` rather than `!isTRUE` as the former will be stricter.
```suggestion
          !(.data$any_flags_hosp) & !(.data$any_flags_ww) &
         isTRUE(.data$ww_sufficient) ~ TRUE,
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+        TRUE ~ FALSE
+      )
+    ) |>
+    dplyr::summarise(
+      n_expected_exclude = sum(.data$ww_expected, na.rm = TRUE)
```
dylanhmorris:
```suggestion
      n_w_ww_expected = sum(.data$ww_expected, na.rm = TRUE)
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+    ) |>
+    dplyr::pull(.data$n_expected_exclude)
```
dylanhmorris:
```suggestion
    dplyr::pull("n_w_ww_expected")
```
Can also use strings in `dplyr::pull` (b/c it's tidy selection, not data masking)

dylanhmorris: For clarity, I'd name the column the same as the quantity that it is going to ultimately represent.

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+  forecast_date_summary_table <-
+    ww_metadata |>
+    dplyr::group_by(forecast_date) |>
+    dplyr::summarize(
+      prop_states_w_ww = sum(ww_data_present) / dplyr::n(),
+      pop_coverage_by_date = sum(pop_coverage * state_pop, na.rm = TRUE) / sum(state_pop),
+      avg_avg_latency = mean(avg_latency, na.rm = TRUE),
+      avg_avg_sampling_freq = mean(avg_sampling_freq, na.rm = TRUE),
+      n_states_w_duplicate_obs = sum(n_duplicate_obs > 0, na.rm = TRUE)
+    )
```
dylanhmorris:
```suggestion
    dplyr::group_by("forecast_date") |>
    dplyr::summarize(
      prop_states_w_ww = sum(.data$ww_data_present) / dplyr::n(),
      pop_coverage_by_date = sum(.data$pop_coverage * .data$state_pop, na.rm = TRUE) / sum(.data$state_pop),
      avg_avg_latency = mean(.data$avg_latency, na.rm = TRUE),
      avg_avg_sampling_freq = mean(.data$avg_sampling_freq, na.rm = TRUE),
      n_states_w_duplicate_obs = sum(.data$n_duplicate_obs > 0, na.rm = TRUE)
    )
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+
+  # Summarize across forecast dates by state
+  state_summary_table <-
+    ww_metadata |>
+    dplyr::group_by(location) |>
```
dylanhmorris:
```suggestion
    dplyr::group_by("location") |>
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,417 @@
...
+    dplyr::summarize(
+      prop_forecast_dates_w_ww = sum(ww_data_present) / dplyr::n(),
+      avg_pop_coverage_by_state = mean(pop_coverage, na.rm = TRUE),
+      avg_avg_latency = mean(avg_latency, na.rm = TRUE),
+      avg_avg_sampling_frequency = mean(avg_sampling_freq, na.rm = TRUE),
+      n_forecast_dates_w_duplicate_obs = sum(n_duplicate_obs > 0, na.rm = TRUE)
+    )
```
dylanhmorris:
```suggestion
    dplyr::summarize(
      prop_forecast_dates_w_ww = sum(.data$ww_data_present) / dplyr::n(),
      avg_pop_coverage_by_state = mean(.data$pop_coverage, na.rm = TRUE),
      avg_avg_latency = mean(.data$avg_latency, na.rm = TRUE),
      avg_avg_sampling_frequency = mean(.data$avg_sampling_freq, na.rm = TRUE),
      n_forecast_dates_w_duplicate_obs = sum(.data$n_duplicate_obs > 0, na.rm = TRUE)
    )
```

dylanhmorris (changes requested):
One major issue concerning computation of sampling frequency and a few tweaks

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: Yeah the word frequency isn't correct here -- its the average days between samples but will give NA if only 1. Also number of days with samples and days total are probably themselves useful metrics. Will add.

dylanhmorris: If you're interested in how big typical gaps are in the data I think still need a different metric. Can discuss.

kaitejohnson: I recomputed as the mean over the sites of (`n_days_w_samples/n_total_days`)

kaitejohnson: The reason being that if you use the whole states data across all sites, this becomes meaningless in states with a large number of sites (they will nearly always have n_days_w_samples = n_total_days so I'm reasoning that this is giving you more information. Also later we average across states and forecast dates in this same way

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: This didn't work, this needs to be `.data$location`. Same with below.  So `group_by()` is also not tidy selection but tidy evaluation  it seems

kaitejohnson:
Confirmed this now produces the expected tables @dylanhmorris

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,423 @@
...
+#' @return ww_metadata: a tibble with a row for each forecast date location and
+#' columns that provide summaries of the wastewater data
+#' @export
+combine_and_summarize_ww_data <- function(forecast_dates,
+                                          locations,
+                                          eval_output_subdir) {
+  if (length(forecast_dates) != length(locations)) {
+    cli::cli_abort(
+      message = c(
+        "Vector of forecast dates and vector of locations must be equal in length",
```
dylanhmorris:
```suggestion
        "Vector of forecast dates and vector of locations must be equal in length. ",
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,423 @@
...
+        "Got {length(forecast_dates)} forecast dates and {length(locations)} locations"
```
dylanhmorris:
```suggestion
        "Got {length(forecast_dates)} forecast dates and {length(locations)} locations."
```

wweval/R/combine_and_summarize_ww_data.R
```diff
@@ -0,0 +1,423 @@
...
+      location = !!this_location,
+      ww_data_present = 1,
+      n_sites,
+      n_labs,
+      pop_coverage,
+      state_pop,
+      avg_latency,
+      avg_sampling_freq,
+      n_days_w_samples,
+      n_duplicate_obs
```
dylanhmorris, line 207: This large block could be a `summarize_ww_metadata` function?

dylanhmorris: Not essential

dylanhmorris (dismissed):
LGTM. A couple optional formatting fixes and suggestions.

wweval/R/combine_and_summarize_ww_data.R
kaitejohnson: I think that is mostly what this is doing! Would rather not break into multiple functions

dylanhmorris (approved):
LGTM

PR #156, opened 2024-09-13
update R version and CI for `cfaforecastrenewalww`

### LINKED ISSUE #154: Expanding compatible R versions
Currently `cfaforecastrenewalww` and `wweval` require `R >= 4.3.0`. Would it be possible to extend this to `R >= 4.1.0`?

kaitejohnson:
Since R-cmd-check installs from github we need to first lower the dependency of `cfaforecastrenewalww` and then we can adjust the dependency on wweval.


dylanhmorris (approved):
LGTM for updating `cfaforecastrenewalww`. One this is merged, prepare a similar PR for `wweval` @kaitejohnson?

PR #160, opened 2024-09-18
Light replacements of cfaforecastrenewalww with wweval and wwinference funs

gvegayon:
This is a minor PR being merged into less-cfaforecastrenewalww that:

1. Adds the `wwinference` package to the list of dependencies.
2. Replaces `cfaforecast::create_dir` with `wwinference::create_dir`.
3. Moves `setup_secrets` from cfaforecast... to `wweval`.
4. Replaces the `cfaforecast::get_params` with `wwinference::get_params`. Notice the `cfaforecast::get_params` returns a data.frame, whereas the `wwinference` returns a list.

All replacing avoids refactoring or any other major changes.


kaitejohnson:
@gvegayon This overall looks good to me. I think the change from params as a dataframe to a list shouldn't affect downstream code.


Would it make sense that on every PR prior to merging, I test a small run in azure and ensure that everything worked as we expect?

gvegayon:
> Would it make sense that on every PR prior to merging, I test a small run in azure and ensure that everything worked as we expect?

That would be awesome!

wweval/R/eval_fit.R
```diff
@@ -23,10 +23,10 @@ eval_fit_ww <- function(config_index,
   }
 
 
-  cfaforecastrenewalww::create_dir(output_dir)
-  cfaforecastrenewalww::create_dir(raw_output_dir)
+  wwinference::create_dir(output_dir)
+  wwinference::create_dir(raw_output_dir)
 
-  params <- cfaforecastrenewalww::get_params(params_path)
+  params <- wwinference::get_params(params_path)
```
kaitejohnson: Ok so  I think the issue is that in the `get_stan_data_list()` function (which we will eventually replace with the one in `wwinference` I do this assigning of the column headers to the parameters in the dataframe, instead of calling them explicitly with $ (bad practice I know). 

So this is failing in the run. My suggestion is not to fix it in `wweval`, but perhaps convert `params` to a dataframe temporarily, since the downstream functionality will eventually be replaced.

kaitejohnson (changes requested):
```
The following objects are masked from ‘package:cfaforecastrenewalww’:

    cleanup_secrets, get_secret, setup_secrets

Rows: 25704 Columns: 4
── Column specification ────────────────────────────────────────────────────────
Delimiter: ","
chr  (1): ABBR
dbl  (2): daily_hosp_admits, pop
date (1): date

ℹ Use `spec()` to retrieve the full column specification for this data.
ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
Rows: 70762 Columns: 4
── Column specification ────────────────────────────────────────────────────────
Delimiter: ","
chr  (1): ABBR
dbl  (2): daily_hosp_admits, pop
date (1): date

ℹ Use `spec()` to retrieve the full column specification for this data.
ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
Prop of population size covered by wastewater: 0.0246778369569984
Error in get_stan_data_list(model_type = "ww", forecast_date = forecast_date,  : 
  object 'uot' not found
Calls: eval_fit_ww -> get_stan_data_list -> get_hosp_data_sizes
In addition: Warning message:
One or more parsing issues, call `problems()` on your data frame for details,
e.g.:
  dat <- vroom(...)
  problems(dat) 
Execution halted
```
Std_err.txt from Azure. 
See my comment where params is loaded in. In order to test your changes locally (now that we have confirmed container build is good), I would run in the command line:
```
Rscript pipeline/command_line_eval_{script_type}_ww.R {config_index}  input/config/eval/eval_config.yaml input/params.toml
```
after making a `eval_config.yaml` file by running `src/setup_eval.R`

wweval/R/eval_fit.R
gvegayon: Just addressed that by adding `|> as.data.frame()` right after all calls to `get_params`. The masking warning will go away once this PR is merged, as the functions mentioned in it are no longer in the `cfaforecast...` package. The warnings show up b/c the package is being installed from main, not from this branch.

kaitejohnson (approved):
Confirmed that the container builds, and local runs of the fit and post process are successful. Think this is good to go!

PR #161, opened 2024-09-18
Additional manuscript analysis

### LINKED ISSUE #165: Add plots of fits compared to calibration and evaluation data in `eval_post_process` for every location-forecast date
## Goal
Make it easy to look at data the models were fit to and evaluated against, so that we can investigate easily individual forecast dates-locations. 

# Tasks 
- [ ] saved image (png) of hospital admissions calibration + evaluation data with fits (median, 50th, 95th percentiles) in nested folder structure along with `input_hosp_data`
- [ ] saved image (png) of site level wastewater concentrations from calibration + evaluation data with fits in nested folder structure along with `input_ww_data`
- [ ] Adding to this the daily growth rate comparison plots described in https://github.com/cdcent/cfa-forecast-renewal-ww/issues/771

kaitejohnson:
This PR creates additional targets to produce supplemental plots and summary tables/metrics. These will be referenced directly in the results and written into the supplement. Outlined analysis planned include but are not limited to:
- [X] plot of CRPS/WIS over time stratified by horizon (CRPS for head to head comparison, WIS for hub comparison)
- [X] summary of overall CRPS/WIS stratified by horizon and overall
- [X] plot of average bias over time for both head-to-head and Hub analysis
- [X] summary stats on the # of forecast_date-locs that ww model were more than X% (10?) better, number that were X% (10?) worse
- [X] summary stats on the number of states where wastewater overall improved forecasts, number of states that overall have worse forecasts, 
- [X] plot of CRPS over time for both hosp only and ww model with fit and forecast comparison figure, for each forecast date and location
- [X] plot of WIS over time for both hosp only and ww model + COVIDhub 4 week ensemble over time + forecast comparison + avg WIS
- [X] plots of the number of sites and population coverage vs forecast performance

See outputs here https://github.com/cdcent/cfa-forecast-renewal-ww/issues/770. I am going to open a separate issue for additional analysis discussed f2f on 2024/10/01

Closes #165 and #168


wweval/R/ms_fig4.R
```diff
@@ -37,6 +37,10 @@ make_fig4_rel_crps_over_time <- function(scores,
 
 
   colors <- plot_components()
+  date_lims <- c(
```
damonbayer: `range(scores$forecast_date)` is equivalent

wweval/R/ms_fig4.R
```diff
@@ -164,6 +168,11 @@ make_fig4_admissions_overall <- function(eval_hosp_data) {
     )
   max_total_hosp <- max(total_hosp$total_hosp)
 
+  date_lims <- c(
```
damonbayer, line 172: Where do these come from?

wweval/R/ms_fig4.R
```diff
@@ -580,9 +592,11 @@ make_fig4_avg_crps_over_time <- function(scores,
     ) +
     theme(axis.title.x = element_blank()) +
     scale_x_date(
-      date_breaks = "2 weeks",
-      labels = scales::date_format("%Y-%m-%d")
+      date_breaks = "1 week",
+      labels = scales::date_format("%Y-%m-%d"),
```
damonbayer: use `date_labels = "%Y-%m-%d"` instead

wweval/R/plots.R
```diff
@@ -183,10 +185,7 @@ get_plot_ww_data_comparison <- function(draws_w_data,
     ) +
     theme_bw() +
     theme(
-      axis.text.x = element_text(
-        size = 8, vjust = 1,
-        hjust = 1, angle = 45
-      ),
+      axis.text.x = element_blank(), # temporarily remove date so can see ww
```
damonbayer: is this "temporary" change meant to be left in?

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
+get_plot_scores_and_forecasts <- function(scores_single_loc_date,
+                                          eval_output_subdir,
+                                          n_calib_days = 10) {
+  this_location <- scores_single_loc_date |>
```
damonbayer: The `pull` with no argument confused me for a bit. `unique(scores_single_loc_date$location)` is more readable in my opinion.

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+    p_scores_t <- ggplot(scores_single_loc_date) +
+      geom_line(aes(x = date, y = crps, color = model),
+        show.legend = FALSE
+      ) +
+      get_plot_theme(
+        x_axis_dates = TRUE
+      ) +
+      scale_x_date(
+        date_breaks = "1 week",
+        labels = scales::date_format("%Y-%m-%d"),
```
damonbayer: use `date_labels = "%Y-%m-%d"`

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+      ) +
+      xlab("") +
```
damonbayer:
```suggestion
      xlab(NULL) +
```

subtly different. `xlab("")` leaves vertical space where the label would go. `xlab(NULL)` erases the vertical space

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+        ),
+        alpha = 0.1
+      ) +
+      geom_ribbon(
+        aes(
+          x = date, ymin = `0.25`, ymax = `0.75`,
+          fill = model_type
+        ),
+        alpha = 0.2,
+      ) +
```
damonbayer, line 119: use [lineribbon](https://mjskay.github.io/ggdist/articles/lineribbon.html) from ggdist?
since each geom uses `x = date`, you can specify `x=date` in the original `ggplot` call and omit it from all the `geom` calls.

wweval/R/supplement_analyses_and_figs.R
damonbayer: This occurs several times. Recommend find and replacing.

wweval/R/supplement_analyses_and_figs.R
damonbayer: This occurs several times. Recommend find and replacing.

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+  )
+}
+
+get_plot_wis_t <- function(hosp_quantiles,
+                           scores,
+                           eval_output_subdir,
+                           submissions_path = "https://raw.githubusercontent.com/reichlab/covid19-forecast-hub/master/data-processed/", # nolint
+                           truth_data_path = "https://media.githubusercontent.com/media/reichlab/covid19-forecast-hub/master/data-truth/truth-Incident%20Hospitalizations.csv", # nolint
+                           hub_comparison_model = "COVIDhub-4_week_ensemble") {
+  truth_data <- truth_data <- readr::read_csv(truth_data_path)
```
damonbayer:
```suggestion
  truth_data <- readr::read_csv(truth_data_path)
```

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+    dplyr::pull()
+
+  loc_code <- cfaforecastrenewalww::loc_abbr_to_flusight_code(this_location)
+
+  this_forecast_date <- hosp_quantiles |>
+    dplyr::distinct(forecast_date) |>
+    dplyr::pull()
+  scores <- scores |> dplyr::filter(
+    forecast_date == !!this_forecast_date,
+    location == !!loc_code
```
damonbayer, line 260: I don't think the `!!` is necessary.

dylanhmorris: Disagree. Additional column names of `scores` aren't known with certainty, so good to evaluate the variables explicitly just in case.

wweval/R/supplement_analyses_and_figs.R
```diff
@@ -0,0 +1,728 @@
...
+
+  colors <- plot_components()
+
+  p <- ggplot(scores_by_horizon_and_t) +
+    geom_line(aes(x = forecast_date, y = mean_score, color = model)) +
+    get_plot_theme(
+      x_axis_dates = TRUE,
+      y_axis_title_size = 8
+    ) +
+    facet_wrap(~horizon, nrow = length(unique(scores$horizon))) +
```
damonbayer: use `ncol=1` instead

wweval/R/ms_fig4.R
```diff
@@ -636,6 +650,9 @@ FGGG
     legend.justification = "left"
   ) #+ plot_annotation(tag_levels = "A") #nolint not working
   fig4
+
+  cfaforecastrenewalww::create_dir(fig_file_dir)
```
dylanhmorris: since we're planning to move away from `cfaforecastrenewalww` wouldn't it be better to use base `dir.create` or `fs::dir_create` here if at all possible? The only potential difference is the directory mode setting done here. Is that still necessary in the context in which you're now running things?

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/8f79e70889c4ca8a1881d84118b3635c6282776e/cfaforecastrenewalww/R/utils.R#L127-L132

wweval/R/supplement_analyses_and_figs.R
dylanhmorris: Either way works for me. I defer to @kaitejohnson here. @damonbayer's base R approach does avoid the need for

```R
dplyr::distinct(.data$location)
```

which is what this will need to be to avoid R CMD check complaints...

dylanhmorris (dismissed):
Agree with @damonbayer's comments expect where noted, but otherwise this looks ready.

wweval/R/ms_fig4.R
damonbayer: followup: can you determine them programmatically instead?

kaitejohnson: Will make them depend on the config

wweval/R/supplement_analyses_and_figs.R
kaitejohnson: Sure, I can change this here.

However, I'll note that this repo is absolutely riddled with the R CMD check complaints...

wweval/R/supplement_analyses_and_figs.R
kaitejohnson: I'm going to push back on this here, I just prefer using `geom_ribbon` explicitly as I have had issues using `lineribbon` with multiple colors and facets. Not saying it wasn't a user error but I prefer to just be verbose and have control than not be able to add all the aesthetic layers I want properly.

damonbayer: I would approve without this change.

wweval/R/ms_fig4.R
kaitejohnson: I will try out using `dir.create()`

kaitejohnson: Yeah I'll go ahead and replace with this throughout

wweval/R/supplement_analyses_and_figs.R
dylanhmorris: Absolutely. But _added_ code should still aim to be compliant.

dylanhmorris (approved):
LGTM

PR #164, opened 2024-09-18
Removing cfaforecastrenewalww package dependency in favor of wwinference

### LINKED ISSUE #147: Scope out necessary steps to use `wwinference` package in eval Azure pipeline
## Goal
The goal of this meta issue is to scope out a refactor of the evaluation pipeline in azure using the [`wwinference` package](https://github.com/CDCgov/ww-inference-model). The package expects a state-level hospital admissions dataset for a single "global" location (jurisdiction) and a site-lab level wastewater data set for the same jurisdiction. Our current eval pipeline uses a `eval_config.yaml` and parses a single index of the vectors of states and forecast dates to run a single model run on a single vintaged dataset. Below we will outline the list of requirements that would be needed to replace the dependency internally to `cfaforecastrenewalww` to use only the package installed from github. 

Requirements:
- [ ] modify `wwinference()` so that it can handle no wastewater data (likely need to pass in empty data and tell it not to be used with `include_ww()` switch, but as is will error without wastewater data). Will require modify `get_stan_data()` (this is being worked on in https://github.com/CDCgov/ww-inference-model/pull/158)
- [ ] edit `setup_container.R` to install from github the latest version of the `wwinference` package, which will be a dependency of the `wweval` package (but likely will need separate instructions).  Compile the model as part of the container build using tooling from `wwinference()`. 
- [ ] remove any dependency on `cfaforecastrenewalww` helper functions e.g. `setup_secrets`. Move these functions to `wweval` if necessary. 
- [ ] modify `eval_fit_hosp.R`  and `eval_fit.R` to use `wwinference()` functions not `wweval` functions, and remove unnecessary/redundant `wweval` functions (e.g. around getting stan data and initialization + make sure that the raw stan object is saved in the current raw output directory in the same format as is expected by `post_process`. 
- [ ] Ensure `post_process` has what it needs to score, generate quantiles, etc. (this might not require anything, but if we modify the input data it might).

gvegayon:
WIP


kaitejohnson:
Closing with #182

PR #167, opened 2024-09-20
165 growth rate comparisons

kaitejohnson:
Figures comparing the growth rate of wastewater data and hospital admissions data using data interpolation and calculation of weekly or daily growth rate. Here I am defaulting to saving weekly as a part of the post-process pipeline. 

See for example https://github.com/cdcent/cfa-forecast-renewal-ww/issues/771


scratch/test_make_growth_rate_plots.R
```diff
@@ -0,0 +1,149 @@
+# Load in DE's 2023-010-16 input data
+library(ggplot2)
+library(zoo)
+input_hosp_data <- readr::read_tsv("output/input_hosp_data.tsv")
```
SamuelBrand1: Is the scratch file supposed to be able to be run by users? That file isn't committed to the repo?

I think its cleaner to either have non-runnable files unavailable or they run.

wweval/R/plots.R
```diff
@@ -1,3 +1,122 @@
...
+#' @export
+get_growth_rate_plot <- function(input_hosp_data,
+                                 input_ww_data,
+                                 location,
+                                 forecast_date,
+                                 rate = "daily") {
+  hosp_data <- input_hosp_data |> dplyr::mutate(
+    admits_7d_rolling = zoo::rollmean(daily_hosp_admits,
+      k = 7, na.pad = TRUE,
+      align = "center"
```
SamuelBrand1: I can imagine wanting to try a different align option (for example in real time analysis). Might be worth making the align a function argument.

SamuelBrand1 (changes requested):
Very neat. Just a couple things.

I'm kind of interested in the difference between the growth rates, but that involves a time shift due to different delays.

scratch/test_make_growth_rate_plots.R
kaitejohnson: Yeah I don't think there's even really a "user" here other than me tbh. I can just delete this file! I find it easier to do some EDA stuff with an example in a script and then convert to a function, so that's all I was doing here

kaitejohnson:
> Very neat. Just a couple things.
> 
> I'm kind of interested in the difference between the growth rates, but that involves a time shift due to different delays.

I was wondering if you think it makes sense to do something more quantitative with this e.g. find the R-squared across a range of lags (e.g. 1-15 days?)

SamuelBrand1:
> > Very neat. Just a couple things.
> > I'm kind of interested in the difference between the growth rates, but that involves a time shift due to different delays.
> 
> I was wondering if you think it makes sense to do something more quantitative with this e.g. find the R-squared across a range of lags (e.g. 1-15 days?)

Yeah, that would be a good idea, but I'm happy to kill this PR off as one that builds some functions. Maybe put a follow up issues or a multi-step analysis plan issue.

scratch/test_make_growth_rate_plots.R
SamuelBrand1: That makes sense, but I was a little sad I couldn't demo by just running!

I think add to gitignore so you can hang onto your test script but I think its overall better not to have code that doesn't run.

Another possibility is to add a function to get some example data; then keep the scripts as examples (maybe add to docs). That would be a new issue for me.

SamuelBrand1 (approved):
LGTM. But in the comments we had some nice follow-up issues.

PR #172, opened 2024-09-25
Use pre-processing from `wwinference` not `wweval`

kaitejohnson:
This isn't functional yet because I (stupidly) created functions with the same name in `wweval` and `wwinference` that do slightly different things and expect slightly different arguments.

In order to get this to work as is, I would have to tweak a lot of the `wweval` functions that will eventually become unnecessary,

I am seeing where we are now and think it is time to rewrite the pre-processing of `wweval` to follow very closely the pre-processing in `wwinference` (so get a NWSS dataset to look like the `ww_data`), and then use all of those functions and delete the ones with duplicate names in `wweval`. This is a bit of a bear of a task and I probably am most suited to doing it since its redundancy that I created...

Tasks:
- [X] rewrite `eval_fit_ww()` and `eval_fit_hosp()` to use the `wwinference` package (currently based on https://github.com/CDCgov/ww-inference-model/pull/158) because we need the swappable `hosp only ` functionality... need to update the parameters here too but for now am using `wwinference` package parameters
- [X] rewite `eval_fit_postprocess()` so that the outputs are compatible with the infrastructure in `wweval` postprocessing. 

Additional tasks that we should do but don't prevent @gvegayon from merging `less-cfaforecastrenewalww-2` are:
- deleting the many unnecessary wweval pre-processing functions #175 
- using the `wwinference()` package to get the draws object instead of the `wweval::get_model_draws_w_data()` #176 

Have made separate issues for these, but in the spirit of making these PRs more manageable going to suggest we do those in separate PRs. 

@dylanhmorris  tagged you for awareness but I think @gvegayon can review this. Basically I just tested and made sure that we can generate the same outputs up until the end of `eval_fit_postprocess()` now using the `wwinference` package. Still a bit clunky but I think we can clean up as we continue to work on this.

Checklist of next steps (as suggested by @gvegayon):
- [ ] use `wwinference::get_draws()` instead of `wweval::get_draws_w_data()` #176 
- [ ] Delete unnecessary pre-processing functions #175 
- [ ] consider rewriting `eval_fit_ww` and `eval_fit_hosp` to just `eval_fit` with an arg indicating whether to fit model with or without wastewater


kaitejohnson:
Ok this is still a WIP, I got up until the point where we are fitting the model with wwinference and almost everything is passing up until the pmfs validation (see here for issue) https://github.com/CDCgov/ww-inference-model/pull/191

wweval/R/eval_post_process.R
```diff
@@ -44,16 +51,15 @@ eval_post_process_ww <- function(config_index,
   eval_hosp_data <- load_object("eval_hosp_data", output_file_suffix)
   input_ww_data <- load_object("input_ww_data", output_file_suffix)
   eval_ww_data <- load_object("eval_ww_data", output_file_suffix)
-  ww_fit_obj <- load_object("ww_fit_obj", output_file_suffix)
+  ww_fit_obj_wwinference <- load_object("ww_fit_obj", output_file_suffix)
+  ww_fit_obj <- ww_fit_obj_wwinference$fit$result
```
kaitejohnson, line 48: I thought this would return what is identical to `ww_fit_obj` above but it doesn't seem to as I can't get the same elements out of it exactly (e.g. flags, diagnostics etc see below).

wweval/R/eval_post_process.R
```diff
@@ -44,16 +51,15 @@ eval_post_process_ww <- function(config_index,
...
+  ww_diagnostics <- ww_fit_obj$diagnostics()
   save_object("ww_diagnostics", output_file_suffix)
-  ww_diagnostic_summary <- ww_fit_obj$summary_diagnostics
+  ww_diagnostic_summary <- ww_fit_obj$diagnostic_summary()
   save_object("ww_diagnostic_summary", output_file_suffix)
-  ww_summary <- ww_fit_obj$summary
-  save_object("ww_summary", output_file_suffix)
   errors <- ww_fit_obj$error
   save_object("errors", output_file_suffix)
   raw_flags <- data.frame(ww_fit_obj$flags)
```
kaitejohnson: Are no flags in our `wwinference` output

wweval/R/get_input_data.R
```diff
@@ -43,51 +43,48 @@ get_input_ww_data <- function(forecast_date_i,
...
-  # If more than one location, than this data isn't being used for fitting
-  # And we don't wanto generate these
-  if (length(location_i) == 1) {
-    site_map <- ww_w_outliers |>
-      distinct(site) |>
-      mutate(site_index = row_number())
-    site_lab_map <- ww_w_outliers |>
-      distinct(lab_wwtp_unique_id) |>
-      mutate(lab_site_index = row_number())
+  deduplicated_data <- ww_data_pkg |>
```
kaitejohnson, line 66: @hannahcohen4

wweval/R/get_input_data.R
```diff
@@ -43,51 +43,48 @@ get_input_ww_data <- function(forecast_date_i,
...
+    dplyr::group_by(lab, site, date) |>
+    dplyr::summarize(
```
damonbayer: good opportunity to use [`dplyr::across`](https://dplyr.tidyverse.org/reference/across.html)

kaitejohnson: I have never used it, can you provide an example?

damonbayer: `summarize(across(c(log_genome_copies_per_ml, log_lod, site_pop), mean))`

is exactly equivalent to your 

```r
dplyr::summarize(
      log_genome_copies_per_ml = mean(log_genome_copies_per_ml),
      log_lod = mean(log_lod),
      site_pop = mean(site_pop)
)
```

wweval/R/eval_fit.R
```diff
@@ -26,7 +26,14 @@ eval_fit_ww <- function(config_index,
   wwinference::create_dir(output_dir)
   wwinference::create_dir(raw_output_dir)
 
-  params <- wwinference::get_params(params_path) |> as.data.frame()
+  # params <- wwinference::get_params(params_path) #nolint
+  # Temporarily get parameters from local package install to test, since
+  # need updates from branch of wwinferece
```
gvegayon: Not following this comment, you may already have a checklist for this; if not, I suggest adding a separate issue for additional fixes needed after this is merged.

wweval/R/eval_fit.R
```diff
@@ -36,46 +43,21 @@ eval_fit_ww <- function(config_index,
...
-  )
   raw_input_hosp_data <- get_input_hosp_data(
     forecast_date_i = forecast_date,
     location_i = location,
     hosp_data_dir = eval_config$hosp_data_dir,
     calibration_time = eval_config$calibration_time
-  )
+  ) |>
+    dplyr::mutate(
+      forecast_date = forecast_date
```
gvegayon: I'm guessing this is not a typo and you are just adding a new column with the same name?

wweval/R/eval_fit.R
```diff
@@ -162,7 +171,14 @@ eval_fit_hosp <- function(config_index,
   wwinference::create_dir(output_dir)
   wwinference::create_dir(raw_output_dir)
 
-  params <- wwinference::get_params(params_path) |> as.data.frame()
+  # params <- wwinference::get_params(params_path) #nolint
+  # Temporarily get parameters from local package install to test, since
+  # need updates from branch of wwinferece
```
gvegayon: Same here about creating an issue with "next steps"/"pending"

wweval/R/eval_fit.R
```diff
@@ -203,42 +202,50 @@ eval_fit_hosp <- function(config_index,
     raw_input_hosp_data = raw_input_hosp_data,
     forecast_date = forecast_date,
     table_of_exclusions = table_of_exclusions
-  )
+  ) |>
+    dplyr::mutate(
+      forecast_date = forecast_date
```
gvegayon: Same here, adding a new variable only?

wweval/R/eval_post_process.R
```diff
@@ -28,7 +28,14 @@ eval_post_process_ww <- function(config_index,
   wwinference::create_dir(raw_output_dir)
 
 
-  params <- wwinference::get_params(params_path) |> as.data.frame()
+  # params <- wwinference::get_params(params_path) #nolint
+  # Temporarily get parameters from local package install to test, since
+  # need updates from branch of wwinferece
```
gvegayon: To the list!

wweval/R/eval_post_process.R
gvegayon: Were you able to debug this?

wweval/R/get_input_data.R
```diff
@@ -266,17 +272,18 @@ clean_ww_data <- function(nwss_subset) {
...
-      ww = pcr_target_avg_conc,
-      ww_pop = population_served
+      site_pop = population_served
     ) |>
     mutate(
       location = toupper(wwtp_jurisdiction),
       site = wwtp_name,
-      lab = lab_id
+      lab = lab_id,
+      log_genome_copies_per_ml = log(pcr_target_avg_conc + 1e-8),
```
gvegayon: Sort of random question, why not using something smaller than that? Why not 1e-20? or why not 1e-10? I don't see the place where this was done previously. There's also `.Machine$double.min.exp` (and others) for you to consider.

gvegayon (commented):
I don't feel I know enough to approve this, but here is my review!

wweval/R/eval_fit.R
dylanhmorris, line 57: I think this is meant to be 
```suggestion
      forecast_date = !!forecast_date
```

dylanhmorris: Seems like this should happen inside `get_input_hosp_data` and should set the `forecast_date` column equal to the value of `forecast_date_i`, no?

wweval/R/get_table_sufficient_ww.R
```diff
@@ -80,9 +80,9 @@ get_ww_data_flags <- function(input_ww_data,
     dplyr::summarize(
       last_date = max(date),
       n_dps = dplyr::n(),
-      prop_below_lod = sum(below_LOD == 1) / dplyr::n(),
+      prop_below_lod = sum(below_lod == 1) / dplyr::n(),
```
damonbayer:
```suggestion
      prop_below_lod = mean(below_lod == 1),
```

wweval/R/eval_fit.R
dylanhmorris: Similar to above it feels like adding the `forecast_date` column belongs inside `get_input_hosp_data`. But also I think DRY could be improved here. Lot of duplication between `eval_fit_hosp` and `eval_fit_ww`.

wweval/R/eval_fit.R
kaitejohnson: Good idea -- I will make a checklist. Basically we need to change the `params.toml` so it has all the same variables the `wwinference` package now expects

wweval/R/eval_fit.R
kaitejohnson: Sure I can make that change!

wweval/R/eval_fit.R
kaitejohnson: The reason for the separate functions is that they used to call different models with different parameters and it was just easier to separate them. Now that this is just a switch, I think a second PR could make this more DRY by just writing a single function `eval_fit` that takes an argument for `include_ww` or not. 

I see this as a separate PR though, with this PR focused on getting the basic functionality working. Will add to the checklist.

kaitejohnson: And in this PR will include in `get_input_hosp_data()`

wweval/R/get_input_data.R
kaitejohnson: Yeah, or I think ideally we would want to pass this as a function arg and set it as a default value. Happy to set that value lower.

I think this wasn't done previously bc the other model took in natural scale concentration values.

wweval/R/eval_fit.R
kaitejohnson: This should be addressed in this PR I think! I replaced the input params with the ones the `winference` package expects. 

Added the checklist for other things to the PR description

wweval/R/eval_post_process.R
kaitejohnson: Yes! It was just a brain fart on my part, I had thought the original `wweval` implementation was returning the raw stan object but it actually is a result of a `sample_model` function that I had written that calculated flags and a few other outputs, and so I added the generation of those outputs to this

wweval/R/eval_fit.R
```diff
@@ -36,46 +36,18 @@ eval_fit_ww <- function(config_index,
   ) |> paste0(".rds")
```
dylanhmorris, line 36: Not for this PR but might be more readable to construct this with `glue::glue`

dylanhmorris (approved):
LGTM!

wweval/R/eval_fit.R
dylanhmorris: happy for this to be in a separate PR, but it should happen.

wweval/R/eval_fit.R
```diff
@@ -94,46 +76,73 @@ eval_fit_ww <- function(config_index,
     calibration_time = eval_config$calibration_time,
     last_hosp_data_date = last_hosp_data_date,
     ww_data_mapping = eval_config$ww_data_mapping
-  )
+  ) |>
+    dplyr::mutate(
+      "location" = !!location,
+      forecast_date = forecast_date
+    )
```
dylanhmorris: Similar to above. Incorporate into the `get_input_ww_data` function in a separate PR.

wweval/R/get_input_data.R
dylanhmorris: Create an issue to discuss how to handle 0s in preprocessing?

dylanhmorris: Given that we have a censoring model it _shouldn't_ ever matter. I think my instinct would be something like 

```R
log_genome_copies_per_ml = ifelse(pcr_target_avg_conf > lod_sewage, log(pcr_target_avg_conc), log(log_sewage / 2)
``` 

or similar

dylanhmorris (commented):
@kaitejohnson will let you resolve the remaining convos and then merge

wweval/R/eval_fit.R
kaitejohnson: I did this in this PR!

PR #177, opened 2024-10-10
176 use wwinference draws

### LINKED ISSUE #176: Use `wwinference::draws()` to get the draws joined with data instead of `wweval::get_draws_w_data()` in `lesscfaforecastrenewalww`
We should use the package functionality we created! And if we need to match for downstream column naming and such, we should just edit after the fact. I think this will be more robust than what we have currently.

kaitejohnson:
This PR closes #176. It does the following:
- Replaces the function to get a draws dataframe that was originally in `wweval` to use the one in `wwinference::get_draws()` for binding the data to the posterior estimates
- Formats the output to match what is expected by the rest of the `wweval` package
- Removes the dependency on `cfaforecastrenewalww` in `wweval`, the container builds, and the pipeline code
- No longer pre-compiles the model in the container build (I couldn't figure out how to do this, since we were previously specifying a relative path within this directory, and now we are using the system file in the package. Currently this works by compiling the model within each VM)

To make sure this works as expected I:
- tested locally 
- tested in the command line
- tested on azure for a single location forecast date

Next steps:
- rerun all location/forecast dates for fit and post process using `wwinference`
- try to run the `targets_eval_postprocessing.R` pipeline, after #179 is also merged. I anticipate there may be a few additional modifications here since a part of this relies on the input data, some of which is now formatted slightly differently (column names differ) . I will change the functions in `wweval` accordingly. The scores, quantiles, etc should be formatted the way that `wweval` expects them downstream.


gvegayon:
> I think this is getting really close to being able to remove cfaforecastrenewalww dependency from wweval/the container build! @gvegayon Do you want me to do that in this PR or do you want to work on in a separate branch merged off of lesscfaforecastrenewalww-2?

This PR is not big, so I think we still have some space to do stuff, so I would say, "go for it, @kaitejohnson!"

kaitejohnson:
Ok will test out trying to remove and see how it goes!

Containerfile
```diff
@@ -12,10 +12,3 @@ ADD . /.
 
 RUN Rscript -e "pak::pkg_install('local::wweval')"
 RUN Rscript -e "pak::pkg_install('argparser')"
-
-RUN mkdir -p stanmodels
-RUN Rscript -e "cfaforecastrenewalww::compile_model( \
-  'cfaforecastrenewalww/inst/stan/renewal_ww_hosp_site_level_inf_dynamics.stan', \
-  'cfaforecastrenewalww/inst/stan', \
-  'stanmodels' \
-)"
```
gvegayon: Not compiling model using `wwinferece` now? Just asking.

setup_container.R
```diff
@@ -30,9 +30,3 @@ pak::pkg_install("github::stan-dev/cmdstanr")
 cmdstanr::install_cmdstan()
 pak::pkg_install("local::wweval")
 pak::pkg_install(additional_deps)
-dir.create("stanmodels")
-wwinference::compile_model(
-  "cfaforecastrenewalww/inst/stan/renewal_ww_hosp_site_level_inf_dynamics.stan",
-  "cfaforecastrenewalww/inst/stan",
-  "stanmodels"
-)
```
gvegayon, line 38: Same q here

wweval/DESCRIPTION
```diff
@@ -89,7 +88,6 @@ Roxygen: list(markdown = TRUE)
 Config/Needs/check: rcmdcheck, testthat
 RoxygenNote: 7.3.2
 Remotes:
-    cfaforecastrenewalww=CDCgov/wastewater-informed-covid-forecasting/cfaforecastrenewalww@less-cfaforecastrenewalww-2,
     stan-dev/cmdstanr,
     wwinference=CDCgov/ww-inference-model
```
gvegayon: Should we point to a release version or something? Right now, it will download the dev version. Do we want that?

gvegayon (approved):
A couple of questions on the bits I better understand. Other than that, LGTM!

Containerfile
kaitejohnson: I tried to using ` system.file("stan", "wwinference.stan",package = "wwinference" )` but I think I might need the full relative file path (this gave me an error when I tried to build the container).

setup_container.R
kaitejohnson: ^ same as above!

wweval/DESCRIPTION
kaitejohnson: I think we should point to a release! I will add

Containerfile
gvegayon: What's the error you are getting?

kaitejohnson: Will try again I didn't save!

kaitejohnson:
```
11/11: RUN Rscript -e "wwinference:compile_model(   'system.file("stan","wwinference.stan", package = "wwinference"),   'system.file("stan",package = "wwinference"),   'stanmodels' )"
.Rprofile for container loaded successfully
Error: unexpected symbol in "wwinference:compile_model(   'system.file(stan,wwinference.stan, package = wwinference),   'system.file"
Execution halted
Error: error building at STEP "RUN Rscript -e "wwinference:compile_model(   'system.file("stan","wwinference.stan", package = "wwinference"),   'system.file("stan",package = "wwinference"),   'stanmodels' )"": error while running runtime: exit status 1
```
I think it is because we can't have R scripting here, just need a file path?

gvegayon: Change

```bash
11/11: RUN Rscript -e "wwinference:compile_model(   'system.file("stan","wwinference.stan", package = "wwinference"),   'system.file("stan",package = "wwinference"),   'stanmodels' )"
```

to 

```bash
11/11: RUN Rscript -e "wwinference:compile_model(   system.file('stan','wwinference.stan', package = 'wwinference'),   system.file('stan', package = 'wwinference'),   'stanmodels')"

```

I think... You have issues with the quotes, that's the problem.

wweval/R/post_processing.R
```diff
@@ -1,48 +1,27 @@
...
-                                   draws,
-                                   forecast_date,
-                                   scenario,
-                                   location,
-                                   input_data,
-                                   eval_data,
-                                   last_hosp_data_date,
-                                   ot,
-                                   forecast_time = 28) {
+new_get_model_draws_w_data <- function(fit_obj_wwinference,
```
dylanhmorris: any reason not to name this the same as the old function?

dylanhmorris (commented):
A couple questions. Concur with @gvegayon's suggestion about quotation marks for the container recipe.

wweval/R/post_processing.R
kaitejohnson: I can change it! Was just comparing while deving

setup_container.R
dylanhmorris, line 1: this shouldn't be needed now that these steps are in the `Containerfile`. Apologies for not noticing previously.

Containerfile
```diff
@@ -11,4 +11,12 @@ RUN Rscript -e "cmdstanr::install_cmdstan()"
 ADD . /.
 
 RUN Rscript -e "pak::pkg_install('local::wweval')"
+RUN Rscript -e "remotes::install_github("CDCgov/ww-inference-model@v0.1.0")
```
dylanhmorris: If this works it's fine, but I'd think you'd want to install these in the reverse order (and/or make `wweval` depend on `ww-inference-model`).

setup_container.R
kaitejohnson: The whole thing isn't needed?

Containerfile
kaitejohnson: `wweval` does depend on `wwinference`, but I still got `wwinference` object not found... this isn't working because of remotes...

kaitejohnson: Ok I think it was just a typo I had!

setup_container.R
dylanhmorris: correct.

dylanhmorris: All the steps that were in `setup_container.R` got moved to the `Containerfile`

dylanhmorris: I am pretty sure I recall deleting`setup_container.R` when I refactored the `Containerfile`, so I am guessing it was reintroduced accidentally in a merge at some point

dylanhmorris (approved):
LGTM

PR #180, opened 2024-10-17
Remove extra functions that aren't documented

kaitejohnson:
Because we are now relying on `wwinference` there is a ton of pre-processing stuff that we don't need that will only make this codebase more confusing/poorly documented. Here's an effort to remove those extra functions involved in pre processing, fitting the model, and immediate postprocessing. 

What done:
- removed `delay_distribs.R`, `pre_processing.R`,  `format_data_for_stan.R`, 'fake_fitting.R` which contained a bunch of not documented functions that were copied over from `cfaforecastrenewalww` and not being used anymore
- fixed typo in `eval_post_process.R` to save plot
- tested changes locally for fit and post process of a single location/forecast date


dylanhmorris (approved):
LGTM

PR #181, opened 2024-10-21
Fix pipeline for targets

kaitejohnson:
This makes some minor tweaks to the functions used in the wweval pipeline so that it expects the outputs from `wwinference`


dylanhmorris (approved):
LGTM

PR #182, opened 2024-10-21
Remove dependency on `cfaforecastrenewalww` and use `wwinference`

kaitejohnson:
This PR completes the removal of any dependence on `cfaforecastrenewalww` in the `wweval` package/pipeline so that the pipeline only uses `wwinference`.


_targets.R
```diff
@@ -249,7 +249,7 @@ list(
   ),
   tar_target(
     name = params_id,
-    command = get_params(param_file_path_id),
+    command = get_params(param_file_path_id) |> as.data.frame(),
```
dylanhmorris: Are these intentional changes? I thought this part of the pipeline been refactored to expect the raw output of `get_params`, but perhaps not

_targets_eval_postprocessing.R
dylanhmorris, line 1: this still has`"cfaforecastrenewalww"` in the `tar_option_set` call. Intentional?

cfaforecastrenewalww/DESCRIPTION
dylanhmorris, line 1: Is there a reason to retain `cfaforecastrenewalww` in `prod` at all?

input/params.toml
dylanhmorris, line 1: Need to adjust some priors in light of changes to the underlying model. Will discuss face to face.

src/generation_interval.R
dylanhmorris, line 1: can this not be removed given that `wwinference` has a default COVID-19 generation interval?

src/inf_to_hosp_delay_distrib.R
dylanhmorris, line 1: Similar comment about delay distribs.

wweval/R/get_input_data.R
```diff
@@ -43,51 +48,67 @@ get_input_ww_data <- function(forecast_date_i,
...
+    all_ww_data,
     scenario_i,
     scenario_dir
   )
-  subsetted_ww_data <- ww_data |>
+  subsetted_ww_data <- all_ww_data |>
     dplyr::filter(wwtp_name %in% !!list_of_site_ids)
 
-  ww <- subsetted_ww_data |>
+  ww_data_pkg <- subsetted_ww_data |>
```
dylanhmorris: I'm guessing this rename to `ww_data_pkg` was to mark it as part of the transition to using `wwinference`. I think we can move back to calling this something like `cleaned_ww_data`. I also think this pipe chain can be longer; not clear to me that `subsetted_ww_data` and `ww_data_pkg` need to be defined as separate variables on the way to `deduplicated_data`

dylanhmorris (changes requested):
Looks very close, thanks @kaitejohnson! A few key questions.

_targets.R
kaitejohnson: Oh no I think these are changes that @gvegayon had made initially, to get things to work, but now are not needed. I will revert (though I don't think we really need to change anything in `_targets.R` since it is not the eval pipeline but the old production pipeline).

_targets_eval_postprocessing.R
kaitejohnson: Nope I don't think its necessary, but will test

cfaforecastrenewalww/DESCRIPTION
kaitejohnson: I don't see why we would delete it, I see it as a nice fall back where we can at least always go back to the production runs from last spring and use it to compare performance.

input/params.toml
kaitejohnson: Oh whoops, I thought I had adjusted these... maybe I just adjusted on azure blob storage which is where they are actually used

src/generation_interval.R
kaitejohnson: Probably! Let me just double check

src/inf_to_hosp_delay_distrib.R
kaitejohnson: Yeah I think all of these are redundant now, though they might have been used in `_targets.R` which might be nice to maintain as a back up production pipeline.

input/params.toml
kaitejohnson: Oh sorry, what you mean here is you want to change the values of the priors. I agree but think we should set up benchmarking first.

src/generation_interval.R
kaitejohnson: So these were identical but we were actually pulling them from the saved pmfs. Eventually, we should pull these from `cfa-param-estimates` but for consistency with last season I think with this analysis we should keep with the package defaults/the ones that we had saved and put into `saved_pmfs`. I replaced with using the package pmfs. I don't necessarily think we should delete these files though, we could just modify them to pull from cfa parameter estimates and then save them ?

dylanhmorris (approved):
LGTM

PR #185, opened 2024-10-22
Issue 184: Fix plots so data below LOD is at LOD

### LINKED ISSUE #184: Set data to the LOD when data is below the LOD
# Goal

In the model, this will be treated as censored data. But for plotting, it would be nice to just have this show up as at the LOD, for both the input data and the evaluation data. This was done manually in the previous pre and post processing so suggest we do the same thing here for post processing.

kaitejohnson:
This PR closes #184. It does the following:
- makes adjustments to post-processing to replace values below the LOD with the LOD
- it modifies the plotting functions run in post-process, which will now label points as below the LOD in red and flagged as outliers in blue
- it also fixes the date alignment on the wastewater draws plot which should help for states with a large number of sites


wweval/R/ms_fig2.R
```diff
@@ -161,6 +161,14 @@ make_fig2_ct <- function(ww_quantiles,
...
       color = "black", show.legend = FALSE
     ) +
+    geom_point(
+      data = quantiles_wide |> filter(.data$below_LOD == 1),
+      aes(x = date, y = log(calib_data)), color = "red", size = 1
+    ) +
+    geom_point(
+      data = quantiles_wide |> filter(.data$flag_as_ww_outlier == 1),
+      aes(x = date, y = log(calib_data)), color = "blue", size = 1
+    ) +
```
dylanhmorris: Use shape as well as color for accessibility?

dylanhmorris: Also, probably more ggplot-ish to handle these cases (for shape and/or color) with an aesthetic mapping rather than manual subsetting.

You could have a column called `observation_status` with values `standard`, `outlier`, `below_lod` and then map `color` and `shape` to `observation_status`.

I realize that here you have fill and/or color mapped to `model` elsewhere, but

A. Since there's no overlap between the discrete values of `model` and the discrete values of this `observation_status` column, you could have a shared manual scale for both.
B. In this particular case, `model` necessarily `== 'ww'`, so it's arguably more readable and ggplotish to do:
```R
model_color <- colors$model_colors$ww
``` 
and then pass that as a parameter (rather than as a mapped aesthetic) to geoms where you want the color/fill to reflect the wastewater model color choice, e.g.
```R
    geom_ribbon(
      aes(
        x = date, ymin = `0.025`, ymax = `0.975`,
      ),
      alpha = 0.2,
      show.legend = FALSE,
      fill = model_color
    ) 
```
C. If you ever do run into a case where you need two different scales for the same aesthetic, you can use [`ggnewscale`](https://eliocamp.github.io/ggnewscale/)

dylanhmorris (commented):
A couple quick ggplot remarks

wweval/R/ms_fig2.R
kaitejohnson: Made those suggested adjustments, I think I'd rather keep the shapes as circles for now if thats ok... if we get comments in review process happy to change

kaitejohnson:
Pending CI passes this is ready for re-review

wweval/R/ms_fig2.R
```diff
@@ -142,45 +142,59 @@ make_fig2_ct <- function(ww_quantiles,
...
     geom_point(aes(x = date, y = log(eval_data)),
       fill = "white", size = 1, shape = 21,
       show.legend = FALSE
     ) +
     geom_point(
-      aes(x = date, y = log(calib_data)),
-      color = "black", show.legend = FALSE
+      aes(
+        x = date, y = log(calib_data),
+        color = observation_status
```
dylanhmorris: Nicely done. Reiterating my last comment that we should also use different shapes in addition to different colors. Can again be done via `shape = observation_status` and `scale_shape_manual`

dylanhmorris (commented):
Reminder about shapes.

kaitejohnson:
Just added shapes

dylanhmorris (approved):
LGTM

PR #187, opened 2024-10-22
Issue 186: Set up preliminary benchmarking

### LINKED ISSUE #186: Set up performance benchmarking
## Goal
We want to be sure that any changes we are making to the model, now via the package, are either rendering forecast performance the same or improving it. This will help us make evidence-based decisions on model changes, including both structural model changes and changes to priors etc. 

## Requirements
- [ ] for the full pipeline run, we should summarize the following outputs: average CRPS across all target horizons (days), forecast dates, locations + average CRPS for each forecast date across all locations and target horizons  + average CRPS for each location across all forecast dates and target horizons
- [ ] for every change to the model, we should benchmark the same outputs on a subset of the data e.g. a few states and a few forecast dates that represent different epidemic phases. This should still run in azure, but it should be quicker and less computationally expensive (and ideally eventually, be automated with self-hosted runners integrated with azure as @gvegayon is working on). 

My current proposal for the first version of this is simply that we track these values in git, similar to what we had set up when the entire pipeline was being run on the VAP. We could just create a folder in `output` called `benchmarking` and set this up, with a table with a tag of the `wwinference` commit hash, version, this commit hash, and the scores as described above. Any model changes would require the subset to run, and in general every time we run the full eval pipeline we would want to keep track of how the performance was changing. It could either be a table that keeps growing or one that we replace each time in a PR (don't have a strong preference here).

Curious others thoughts on this @dylanhmorris @seabbs @damonbayer @gvegayon 

I would propose we do this before tweaking priors on current model, as well as consider merging into `old-prod` (aka yesterday's prod) to see how our performance shifted before using `wwinference` (and whether we should seriously reconsider a restructure back to the original implementation).

kaitejohnson:
WIP
This PR closes #186. What it does:
- Sets up performance benchmarking for all forecast date locations in the evaluation pipeline (`all_forecasts.yaml`), designed to run in targets. Current results from `wwinference@v0.1.0` are included here
- Set up a targets pipeline and create an eval config to benchmark only a subset of locations/forecast dates. I chose these somewhat vibes based, feel free to leave alternative suggestions
- manually include `wwinference` version.

Open question:
- Is there a way to automate adding the commit hash/version of `wwinference` to this metadata. @dylanhmorris Do you know a good way to do this? I guess I could parse through the `DESCRIPTION` file in `wweval`?


_targets_subset_benchmarking.R
```diff
@@ -0,0 +1,102 @@
...
+    packages = c(
+      "cmdstanr",
+      "rlang",
+      "tibble",
+      "ggplot2",
+      "dplyr",
+      "lubridate",
+      "cmdstanr",
+      "tidybayes",
+      "cfaforecastrenewalww",
```
dylanhmorris: This presumably shouldn't be here?

dylanhmorris: To avoid silent deps you might want to uninstall `cfaforecastrenewalww` locally.

output/benchmarking/subset_forecasts.yaml
dylanhmorris, line 1: I think it makes more sense to have this information in a tabular format rather than as a `yaml` file. Do you disagree @kaitejohnson?

kaitejohnson: I have multiple tables within this yaml. I think I want the three tables and the metadata, and this seemed like the most logical way to store that?

dylanhmorris: Four individual files isn't that many, and `yaml` doesn't make row-wise data very human-readable. I also think `overall_scores` and `scores_by_location` could be a single table, where "all" is a location

kaitejohnson: Would your suggestion be tsvs, named by `benchmark_scope` and then either by location or forecast date, with columns for the two hashes/versioning?

dylanhmorris:
> Is there a way to automate adding the commit hash/version of wwinference to this metadata. @dylanhmorris Do you know a good way to do this? I guess I could parse through the DESCRIPTION file in wweval?

Could use the `package_info()` function from the `sessioninfo` package:

Try

```R
sessioninfo::package_info("wwinference", dependencies=FALSE)$source
sessioninfo::package_info("wwinference", dependencies=FALSE)$ondiskversion
sessioninfo::package_info("wweval", dependencies=FALSE)$source
sessioninfo::package_info("wweval", dependencies=FALSE)$ondiskversion
```

output/benchmarking/subset_forecasts.yaml
dylanhmorris: Yes, that's one option. One or two columns with duplication don't bother me too much.

Alternatively/additionally, could have:
- versioning in the filenames
- a benchmark run id in the filename and a lookup table of versions associated to the run ids

kaitejohnson: I think I have slight preference for the filename being constant. Will modify to the tabular format

kaitejohnson:
It isn't the prettiest logic but heres the latest and stacked tables

wweval/R/benchmarking.R
```diff
@@ -84,34 +85,96 @@ benchmark_performance <- function(ww_scores,
...
+      )
+    ))) {
+      df <- readr::read_tsv(file.path(
+        benchmark_dir,
+        glue::glue(
+          "{benchmark_scope}_by_forecast_date.tsv"
+        )
+      ))
+      # check that wwinference hash is different
+      if (df$wwinference_version[1] != wwinference_version) {
```
dylanhmorris: also check `wweval` version here?

wweval/R/benchmarking.R
dylanhmorris, line 1: Add a quick unit test for the appending/overwriting on temporary files with fake benchmark data?

dylanhmorris (changes requested):
Very close now. Two final things.

wweval/tests/testthat/test_benchmarking.R
```diff
@@ -0,0 +1,45 @@
...
+  location = c("x", "y"),
+  forecast_date = rep("2023-10-15", 2),
+  model = rep("hosp", 2),
+  crps = c(0.5, 0.6),
+  bias = c(4, 5),
+  ae_median = c(0.3, 0.2)
+)
+
+test_that("benchmarking writes files correctly", {
+  # Assume stan_models_dir is a directory containing your .stan files
```
dylanhmorris:
```suggestion
```

dylanhmorris (commented):
LGTM 

EDIT: saw you already removed

dylanhmorris (approved):

Issue #188, opened 2024-10-23
Test out modifying infection feedback prior


kaitejohnson:
## Goal
We have a suspicion that the infection feedback prior was way too high. We are in the process of setting up benchmarking (#187 ) and we should set up a run to compare the performance across all forecast dates and locations with the new prior and use that as a basis for including this in a new version of `wwinference`.

PR #189, opened 2024-10-23
Issue 188: Benchmark infection feedback performance

kaitejohnson:
This PR addresses #188 and #194. It contains the outputs for `v0.1.0` of the `wwinference-version`, `227-inf-feedback` which uses a diffuse prior of `logmean(0)` and logsd(3), and lastly, `227-inf-feedback-posterior-mod-eta` which uses the posterior from the longer fits of the subset to assign the priors for the 90 day fits across a few forecast dates and locations (so inf_feedback logmean = log(90), logsd = log(1.9)) and R(t) RW step size mean = 0.0278, sd = 0.01. 

My only consideration would be if, based on the discussion below, we would justify a tighter prior on infection feedback (which might allow for improved performance at the peak). I think based both on the approximation from @SamuelBrand1 and the tuning based on a longer fit, we have justification for an infection feedback in this regime. 

Here are the visual results of the subsetting comparing across the 3 cases. Overall, current version still looks better but I think we should 1 run on all forecast dates-locs and 2 consider a slightly tigher prior
subset_forecasts_by_forecast_date
subset_forecasts_by_location
subset_forecasts_overall

Additional tasks done in this PR:
- sets up `output/benchmarking/plots` to compare visually the tables in the scores by forecast date and location.


output/benchmarking/latest_subset_forecasts_by_location.tsv
```diff
@@ -1,7 +1,7 @@
...
-7.110957447979565	7.109562940863587	0.10033991228070176	0.18538004385964912	9.346454236343268	9.529331764656112	all	d23b28a	v0.1.0	2024-10-24T17:48:21Z
+1.0376521512900907	1.0118669366748054	-0.036991228070175436	-0.007656798245614037	1.4245526802625763	1.4264265157766278	AK	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
+24.626800317489625	19.49484775911998	0.2060657894736842	0.4117872807017544	30.689131156635455	24.783400147179314	MA	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
+2.9839338399865656	1.8872799196790222	-0.01010087719298246	0.04486293859649123	4.248055546108953	2.2765310519617756	NH	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
+50.50136383077716	41.39684785751621	0.1817642543859649	0.2782510964912281	62.26077007486285	51.60308220739496	NJ	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
+4.468227694679483	10.723086482359754	-0.015425438596491228	0.33135855263157893	6.534705999620968	14.015772038017861	WA	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
+16.723595566844583	14.902785791069954	0.0650625	0.2117206140350877	21.03144309149816	18.821042392066108	all	51ddee5	227-inf-feedback	2024-10-25T17:16:51Z
```
kaitejohnson: Looks to me like things are particularly worse in NJ, MA, WA on 10-16 and 2024-01-08. Will make a bunch of comparison figures in a cdcent/cfa-forecast-renewal-ww

input/params.toml
```diff
@@ -42,8 +42,8 @@ offset_ref_initial_exp_growth_rate_prior_sd = 0.025
 autoreg_p_hosp_a = 1 # shape1 parameter of autoreg term on IHR(t) trend
 autoreg_p_hosp_b = 100 # shape2 parameter of autoreg term on IHR(t) trend
 eta_sd_sd = 0.01
-infection_feedback_prior_logmean = 6.37408 # log(mode) + q^2 mode = 500, q = 0.4
-infection_feedback_prior_logsd = 0.4
+infection_feedback_prior_logmean = 0 # log(mode) + q^2 mode = 1, q = 3
+infection_feedback_prior_logsd = 3
```
SamuelBrand1: I'm probably not understanding something, but is this is a median = 1 and `std = 8102.583912148207` prior replacing a  `median = 586.4456526813148` and `std = 264.62731766803165` prior. E.g standard parameterisation of Log-Normal?

SamuelBrand1: I.e. should we be interpreting this as a comparison between a confident high prior vs an uninformed prior?

kaitejohnson: Yes and since the 90 day window probably doesn't let the model learn infection feedback much (particularly near the peak when its mostly been growing), I suspect we will need a stronger prior (e.g. more informative but lower, next thing @dylanhmorris and I discussed trying is `logmean = log(100)` , `logsd = log(3)` so `median = 100`, `stdev = 279`

SamuelBrand1:
Hey @kaitejohnson @dylanhmorris I've been thinking about appropriate priors for $\gamma$ (the feedback term).

The essential modelling idea is that high levels of incidence rate (e.g. observed hosp) are associated with the _effective_ reproduction number $R(t)$ decreasing towards 1, whilst remaining agnostic on the cause (described [here](https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/prod/model_definition.md#instantaneous-reproduction-number)).

If we rewrite the feedback term in a non-dimensional way whilst linked to the observational data:

```math
\gamma = \hat{\gamma} \frac{p_{\text{hosp}}}{N}.
```

Then, looking at the approximate hospitalisation density, in per capita terms, $H_{\text{max}}$ when epidemic turn-overs occur, i.e. when the effective reproductive number goes to 1 we get the following approximation at the peak time:

```math
\begin{aligned}
R^u \exp\left( - \gamma \langle I \rangle_g \right) & \approx R^u \exp\left( - \hat{\gamma}H_{\text{max}} \right) = 1. \\
\implies & \hat{\gamma} = \frac{ \log\left( R^u \right)}{H_{\text{max}} }
\end{aligned}
```
Where $\langle I \rangle_g$ is the generation distribution weighted average of (true) incidence at the peak.

This allows us to have a reasonable informed prior, using Covid as an example:

- My understanding is that in the last couple of seasons the peak hospitalisation rate https://covid.cdc.gov/covid-data-tracker/#covidnet-hospitalization-network is around 5-10 per 100,000 people, i.e. a density number of `5e-5` - `1e-4`.
- $p_{\text{hosp}}$ is around `1e-3` to `1e-2` (I think).
- $R^u$ is around `1e1`?

This suggests that the non-dimensionalised feedback has the order of magnitude $\hat{\gamma} \sim \mathcal{O}(10^4)$. This suggests that our _a priori_ feedback magnitude for a state of population size $N$ should be about:

```math
N \gamma \sim \mathcal{O}(10^2).
```

E.g for CA (about 40 million people) maybe a prior of about `LogNormal(logmean = log(1e-5), logsd = 1.0)` with the `logsd` reasoning being based on the accuracy of the reasoning enough being around an order of magnitude.

input/params.toml
SamuelBrand1: >  I discussed trying is `logmean = log(100)` , `logsd = log(3)` 

I think this agrees with my reasoning below on the $N \gamma$ prior and makes sense.

kaitejohnson:
@SamuelBrand1 Thanks for this -- I am going to try to walk through how I would logic through this as well with the assumption that both $I(t)$ and $H(t)$ are per capita. 

The infection feedback term impacts $R(t)$ as follows:
```math
R(t) = R^u(t) exp(-\gamma \sum I(t-\tau) g(\tau))
```
We can write $I(t-\tau)$ as approximately (ignoring delays for now) $I(t-\tau) \approx \frac{H(t)}{p_{hosp}}$. Can then define 
```math
\hat{\gamma} = \frac{\gamma}{p_{hosp}}
``` 

Then assuming infection when R(t) goes to 1 is around when hospitalizations peak we can write:

```math
1 = R^u(t) exp(\gamma \sum \frac{H(t-\tau)}{p_{hosp}} g(\tau)) = R^u(t) exp(\hat{\gamma} H_{max})
```

Solving for $\hat{\gamma}$ we get:
```math
\hat{\gamma} = \frac{log(R^u)}{H_{max}}
```
and then for $\gamma$

```math
\gamma = \frac{log(R^u) p_{hosp}}{H_{max}}
```

I would assume $R^u$ closer to `1.5-2`,  peak hospitalizations around `1e-4` to `5e-5` and IHR around `1e-2` gives somewhere between `40-140` roughly so perhaps logmean = log(90), logsd ~ log(0.5), with the option to make logsd larger to make it slightly less informative?

scratch/get_empirical_posterior_vals.R
```diff
@@ -0,0 +1,71 @@
...
+## eta_sd---------------------------------------------------------------
+mean_eta_sd <- mean(eta_sd_draws$eta_sd)
+sd_eta_sd <- sd(eta_sd_draws$eta_sd)
+
+message("Empirical mean of RW step size across 5 locations: ", mean_eta_sd)
+message("Empirical sd of RW step size across 5 locations: ", sd_eta_sd)
+
+## inf_feedback----------------------------------------------------------
+mean_inf_feedback <- mean(inf_feedback_draws$infection_feedback)
+sd_inf_feedback <- sd(inf_feedback_draws$infection_feedback)
```
dylanhmorris: Better to compute `mean(log(inf inf_feedback_draws$infection_feedback))` directly and likewise for `sd(log(inf_feedback_draws$infection_feedback)`

scratch/get_empirical_posterior_vals.R
```diff
@@ -0,0 +1,71 @@
...
+)
+
+message(
+  "Empirical logmean of infection feedback across 5 locations: ",
+  logmean_inf_feedback
+)
+message(
+  "Empirical logsd of infection feedback across 5 locations: ",
+  logsd_inf_feedback
+)
```
dylanhmorris, line 64: Any reason to print to screen rather than writing to disk?

dylanhmorris (changes requested):
One technical point.

scratch/get_empirical_posterior_vals.R
kaitejohnson: Happy to

scratch/get_empirical_posterior_vals.R
kaitejohnson: Ok, either way I think the values are approx anyway so I don't think this negates the validity of the current run?

dylanhmorris: Agree; mostly for going forward.

SamuelBrand1:
> I would assume R u closer to `1.5-2`, peak hospitalizations around `1e-4` to `5e-5` and IHR around `1e-2` gives somewhere between `40-140` roughly so perhaps logmean = log(90), logsd ~ log(0.5), with the option to make logsd larger to make it slightly less informative?

Yes, this makes sense. I was a bit confused because I thought $I(t)$ was modelled as population numbers, but now I see that its population density (per capita) with $N$ coming in when comparing to numbers data. This all makes sense to me now.

dylanhmorris (approved):

SamuelBrand1:
NB: on a f2f @kaitejohnson I noticed that my reasoning above was based on _weekly_ hosp rates so I have to divide that by 7 to make the correct scale estimator.

Fortunately, @kaitejohnson was smart enough to fix this anyway!

kaitejohnson:
Not exactly, but it basically cancels out because of our differences in assumptions about $R^u(t)$

PR #190, opened 2024-10-24
Hot fix add csvs

kaitejohnson:
This PR reverts some of the changes made in haste to the set up of writing the eval config, returning to the original structure where the local file structure is not time stamped (instead we will time stamp in blob storage via the batch_config and then fuse the folder to local (or what I am doing now which is just downloading it)). Don't have unlimited local storage so I dont think we want vintages locally!

Also since making the changes to `benchmarking.R` needed to rerun for the `wwinference @v0.1.0` performance to generate csvs for the subset and all forecast date locations


src/setup_subset_benchmarking.R
```diff
@@ -28,5 +28,6 @@ write_eval_config(
   # score for all locations one forecast date
   overwrite_benchmark = TRUE, # Set as TRUE if want to save outputs of
   # benchmarking in directory,
-  name_of_config = "benchmark_config"
+  name_of_config = "benchmark_config",
+  wwinference_version = "v0.1.0"
```
dylanhmorris, line 32: Use the hash here?

wweval/R/benchmarking.R
```diff
@@ -100,101 +100,119 @@ benchmark_performance <- function(ww_scores,
...
+          forecast_date = readr::col_date(),
+          crps_hosp = readr::col_double(),
+          crps_ww = readr::col_double(),
+          bias_hosp = readr::col_double(),
+          bias_ww = readr::col_double(),
+          ae_hosp = readr::col_double(),
+          ae_ww = readr::col_double(),
+          wweval_commit_hash = readr::col_character(),
+          wwinference_version = readr::col_character(),
+          time_stamp = readr::col_character()
```
dylanhmorris: `col_datetime`?

wweval/R/benchmarking.R
```diff
@@ -100,101 +100,119 @@ benchmark_performance <- function(ww_scores,
...
+          crps_hosp = readr::col_double(),
+          crps_ww = readr::col_double(),
+          bias_hosp = readr::col_double(),
+          bias_ww = readr::col_double(),
+          ae_hosp = readr::col_double(),
+          ae_ww = readr::col_double(),
+          location = readr::col_character(),
+          wweval_commit_hash = readr::col_character(),
+          wwinference_version = readr::col_character(),
+          time_stamp = readr::col_character()
```
dylanhmorris: again, `col_datetime`?

dylanhmorris (commented):
Looks very good. A couple very minor questions

src/setup_subset_benchmarking.R
kaitejohnson: I made this so that its an option to either specify or use the hash-- these results were run using the v0.1.0 (even though that is not currently what is installed locally) and so I want to just specify manually here.

dylanhmorris (approved):
LGTM

PR #198, opened 2024-10-30
Modify eta and inf feedback

kaitejohnson:
This is based on the logmean and logsd of the posteriors run for 365 days for the 5 example states, figures here https://github.com/cdcent/cfa-forecast-renewal-ww/issues/774.

These posteriors from the above PR were used to estimate the infection feedback and R(t) RW step size, acknowledging that we are now using our training dataset on our testing data set (fitting to the fill forecast period and then comparing performance), basically tuning the model. 

These are logmean = log(90), logsd = log(1.89) for infection feedback. For eta_sd the mean was 0.278, which for now we will use as the sd (can change so mode is not 0 if this works well)


kaitejohnson:
Plots of results compared to previous priors we have tried, green is this latest run. I think the overall results are pretty driven by the forecasts from 2024-01-08, so the relative rankings might change if  we were to have run all the forecast dates for all locations. 

subset_forecasts_by_forecast_date
subset_forecasts_by_location
subset_forecasts_overall

I am going to go ahead and do another run where I change the prior mode of eta_sd in the `wwinference` stan model, since this was just changing the prior on the stdev in the RW step size with a mode still at 0.

kaitejohnson:
With the eta_sd mean modified

PR #204, opened 2024-11-02
203 update scoringutils 2.0.0

### LINKED ISSUE #203: Use new `scoringutils 2.0.0` in pipeline
Updated `scoringutils` on cran introduces breaking changes to the scoring as implemented in the earlier version. Rather than rely on the deprecated version in `wweval`, we should fix breaking changes using the updated version.

kaitejohnson:
This PR closes #203. It uses `v2.0.0` of `scoringutils` to compute the full set of scores based on the samples and quantiles of the forecasted hospital admissions in the evaluation pipeline.

Note this is a WIP. Currently we have just edited the inputs to `scoringutils::score()` which expects new names for the predicted and observed values and requires converting from a `data.table` to a forecast object. For compatibility with the rest of the downstream functions (mostly plotting) I converted names back to what we had previously named the columns (`true_value` and `prediction`). This might be incompatible with `summarise_scores()`. 

To do 
- [ ] test `eval_postprocess()` locally on a `ww_fit_obj` and ensure the scores are equivalent or nearly to the previous implementation
- [ ] run on full `_targets_eval_postprocessing.R` or `targets_subset_benchmarking.R` to ensure compatibility with downstream `scoringutils` functions


kaitejohnson:
Going to close as not planned for now

PR #212, opened 2024-11-26
Issue 211: Add real-time hosp only model to analysis 

### LINKED ISSUE #211: Add hosp only to the real-time analysis in Hub comparison figure
Ideal state is to have the top row of current figure 5 duplicated, with one row for the real-time period where we actually submitted, and another for the entire season using the retro models for comparison

kaitejohnson:
This PR closes #211. 
What it does:
- adds to the `score_real_time_outputs()` function to allow the option to score the quantiles with the wis. Note this makes strong assumptions that are particular to this analysis (e.g. when we use wis, we are only scoring forecasts, not nowcasts so the quantiles should be truncated as an additional step).
- adds a function that formats the wis scores from both models, and filters to the hosp only model wis (we already have the real-time scores from github and I think we should maintain this integrity -- just in case what we have saved in our file system is not identical to the submission)
- adds to the targets pipeline a step where the `cfa-hosponlyrenewal(real-time)` score tibble is bound to the other real time scores
- adds a plotting color for the real-time hosp only model
- creates a new Hub comparison [figure ](https://github.com/cdcent/cfa-forecast-renewal-ww/issues/778#issuecomment-2504412601)where we split it such that the top two rows are figures from the Feb- Mar real-time period, and the bottom two rows are from the entire season and include the retrospective versions of the models (see here for a draft, note I spent way too long tinkering with formatting of legends and decided to just remove for now and can add them in via ppt in final draft...)
- adds an entirely new figure for the real-time relative performance to mock the retrospective relative performance figure, but using the WIS not the CRPS 
- computes the relative scores always at the level of an individual forecast, averaging across all horizon days
- any summary of the distribution of relative crps scores now uses the geometric mean, and no longer improperly truncates the data before computing the geometric mean


wweval/R/ms_fig3.R
```diff
@@ -319,11 +316,12 @@ make_fig3_crps_underlay_fig <- function(scores,
...
     get_plot_theme(
       x_axis_dates = TRUE,
-      y_axis_title_size = 8
+      y_axis_title_size = 8,
+      y_axis_text_size = 6
     ) +
     scale_y_continuous(
       # don't expand y scale at the lower end
-      expand = expansion(mult = c(0, 0.05))
+      limits = c(0, max_crps + 0.05)
```
dylanhmorris: This is the problem. y-axis truncation should happen in `coord_cartesian`, not `scale_y_continuous`, unless you actually want to truncate the data (but I don't think you do).

kaitejohnson: Open Q whether we want these to have a shared y-axis or not -- the current version of the manuscript has a "free" y axis for each horizon, but I think it might be reasonable to consider making them all standardized.

Otherwise I think we can just get rid of this.

kaitejohnson:
@dylanhmorris I am going to spend some time cleaning this up tomorrow to make the sections clearer and align with current figures (will try to get rid of numbers in case we move stuff..), and then I think it is ready for your review.

kaitejohnson:
Going to go ahead and merge this and just have one PR for review

PR #220, opened 2024-12-13
add bar chart as main fig and swap into Hub comparison fig

kaitejohnson:
This PR swaps out a supp fig for a main fig and documents it.

Update: 
- removes CDC-Covidhub ensmble because it is redundant
- makes absolute heatmaps have a single continuous color scale


dylanhmorris (approved):
Where's here LGTM, thanks @kaitejohnson! Will let you merge.

PR #221, opened 2024-12-15
add supp fig, add asterisk to fig4

kaitejohnson:
Minor fixes to address recent changes:
- add asterisk to the name of hosp only model in real time 
- save a version of the wis over time figure for all the hub models


dylanhmorris (approved):
LGTM

PR #227, opened 2025-02-02
Analysis plan for investigation into the heterogeneity in wastewater forecast performance

kaitejohnson:
EDITED:

This document outlines the planned analyses (with references to results removed) to address the requests from internal reviewers. This plan describes four  "hypothesis-generating" analyses using the scores from the forecasts and the raw data used as input into the forecasts, intended to answer two broad questions about why the wastewater-informed versus hospital admissions only models diverge in their predictions when they do, and what, if anything, could predict whether the models will make distinct predictions and which will perform better when they do. The goal of these analyses are to address questions around performance heterogeneity with respect to including wastewater into _our specific model_, and should as always, not be extrapolated to wastewater-informed forecasting beyond this context. 

The four analyses are as follows:
- strength-of-trend analyses
- absolute versus relative performance analyses
- score component analyses
- spurious trend in wastewater/investigation into impact of using flow-pop normalized concentrations

We would like to merge the analysis plan into the main codebase in order to "pre-register" the additional analyses to be performed prior to resubmitting to clearance. 

Tagging @dylanhmorris and @ztm5 for awareness.


dylanhmorris (approved):
Thanks @kaitejohnson!

kaitejohnson:
@dylanhmorris not sure why CI is now failing

kaitejohnson:
Closing per requested

PR #230, opened 2025-02-03
Update batch workflow to use `azuretools`

dylanhmorris:
This PR:

- Removes the need for a `batch_config.toml`, provided the user has appropriately configured environment variables.
- Allows us to use updated service principals, batch accounts, and runner images, avoiding deprecation warnings and increasing available compute.
- Updates the batch README to reflect this, and expands/edits it for clarity.


src/write_eval_config.R
```diff
@@ -132,13 +132,6 @@ write_eval_config <- function(locations, forecast_dates,
     as.vector() |>
     unique()
 
-  # Table of run ids for accessing real-time model fits
-  real_time_output_dir <- file.path("output", "real_time_outputs")
-  path_to_table_of_run_ids <- file.path("output", "real_time_outputs", "table_of_run_ids.rds")
-  table_of_run_ids <- readRDS(path_to_table_of_run_ids)
-  table_of_run_ids$forecast_date <- as.character(table_of_run_ids$forecast_date)
```
dylanhmorris, line 139: @kaitejohnson This fails unless the user has these (private) outputs locally. I am guessing these are needed for the hub comparison analysis. AFAICT they are not needed for the actual batch jobs. Is that correct? If so, perhaps simplest to put the configuration for them elsewhere, e.g. in the relevant targets pipeline, and add some README specification about where to obtain those results if you don't get have them.

kaitejohnson: @dylanhmorris They are not needed for batch no, but the `eval_config.yaml` is for the entire pipeline to generate all the manuscript figures, not just for batch. I see your point as this can't fail for someone else to run the batch jobs, but I'd prefer we just add an if else and set the default to FALSE for these components than do any sort of refactor of the configuration because:
 1. I can't test anything without access 
 2.  Even when I have access, this does not strike me as necessary towards finishing the paper in the very limited time I have to work on this.

dylanhmorris: > @dylanhmorris They are not needed for batch no, but the eval_config.yaml is for the entire pipeline to generate all the manuscript figures, not just for batch

Thanks! That's what I figured but wanted to confim.

> I'd prefer we just add an if else

Implemented in c9c99a1

batch/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Set up your environmental variables
+We'll use the `EnvCredentialHandler` from the [`azuretools`](https://github.com/CDCgov/cfa-azuretools) Python library to handle credentials for CFA Azure resources. It looks for key configuration in your environment variables. CFA's STF Team provide a secret-free (but private) `azureconfig.sh` script to configure environment variables appropriately in their [SharePoint](https://cdc.sharepoint.com/:u:/r/teams/CenterforForecastingandOutbreakAnalytics/Shared%20Documents/General/02%20-%20Predict/Real%20Time%20Monitoring%20(RTM)%20Branch/Short%20Term%20Forecasts/azure/azureconfig.sh?csf=1&web=1&e=e7YBqr). Contact @dylanhmorris if you believe you should have access and do not. We recommend setting up those environment variables as part of your terminal setup, e.g. by adding the following to your [`.bash_profile`](https://linuxopsys.com/dotfiles-in-linux-explained):
```
kaitejohnson: I have never edited by `.bash_profile`, is this also in the home directory?

batch/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+```bash
+. <path to your azuretools.sh from your home directory>
+```
+
+Using the values of the environment variables defined in `azureconfig.sh`, `EnvCredentialHandler` should be able to retrieve a set of valid credentials on your behalf from an Azure Key Vault, provided you have run `az login`.
+
```
kaitejohnson, line 185: My understanding is that this replaces the need for clicking on the url in the terminal and opening the browser when running `az login`?

batch/README.md
```diff
@@ -0,0 +1,222 @@
...
+In our example, `setup_job.py` creates a bunch of tasks, all of them consist of running the following command for different values of {config_index}:
+```
+Rscript pipeline/command_line_eval_{script_type}_ww.R {config_index}  input/config/eval/eval_config.yaml input/params.toml
+```
+Each invocation of that command will perform one of the model fits specified in `eval_config.yaml`; which one depends on the value of `{config_index}`. `setup_job.py` loops over possible values of `{config_index}`, creating tasks for each one.
+
+Let's run `setup_job.py` to create a model fitting job and its constituent tasks. We'll name it `my-demo-fit-job` and have it run on the `wastewater-demo-pool` we just created. Let's image we have a properly formatted configuration file named `eval_config.yaml` stored in `input/eval_config.yaml`. You can create one using the [`src/setup_eval.R`](../src/setup_eval.R) R script.
```
kaitejohnson:
```suggestion
Let's run `setup_job.py` to create a model fitting job and its constituent tasks. We'll name it `my-demo-fit-job` and have it run on the `wastewater-demo-pool` we just created. Let's image we have a properly formatted configuration file named `eval_config.yaml` stored in `input/config/eval/eval_config.yaml`. You can create one using the [`src/setup_eval.R`](../src/setup_eval.R) R script. Note this must be the same config file you uploaded to blob storage in azure storage container for the wastewater input.
```

batch/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+```bash
+python3 batch/setup_job.py input/eval_config.yaml fit my-demo-fit-job wastewater-demo-pool
+```
```
kaitejohnson: So the difference between this and the previous version is that the job name and the pool name are command line input arguments instead of accessed from the `batch_config.toml`?

batch/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+This should create a job named `my-demo-fit-job` consisting of tasks that are named by forecast dates, locations, scenarios and the the job type. (here `fit`). Once your fitting job is finished, set up a second job to postprocess it by running:
+
+```bash
+python3 batch/setup_job.py input/eval_config.yaml post_process my-demo-postprocess-job wastewater-demo-pool
```
kaitejohnson:
```suggestion
python3 batch/setup_job.py input/config/eval/eval_config.yaml post_process my-demo-postprocess-job wastewater-demo-pool
```

kaitejohnson (commented):
A few questions. Also my slight preference would for you to walk through an example with this codebase so I can see that it runs as expected, since I can't test it yet

batch/README.md
dylanhmorris: Yes it is. `~/.bash_profile`. The [link](https://linuxopsys.com/dotfiles-in-linux-explained) gives a good quick overview.

batch/README.md
dylanhmorris: This replaces the need for the manual retrieval of credentials we had before (and the need to specify where to find them in a `batch_config.toml`, e.g. it replaces things like [this](https://github.com/cdcent/cfa-forecast-renewal-ww/blob/4b1de70bd095b9c3a942de0fb8c230de7d896524/batch/setup_job.py#L53-L64)

You'll still need to run `az login`. I will rewrite to make that clearer.

batch/README.md
dylanhmorris: Correct, which makes it easier to make new pools and jobs without having configs proliferate or get modified in place.

batch/README.md
kaitejohnson: No that was clear originally, it just used to require you clicking on a link -- is that still the case?

dylanhmorris: The link thing is on the `az login` side, so that will still happen (or a browser will open automatically if you have a browser set up in wsl). Should we document that here?

kaitejohnson: Hmm yeah perhaps would be useful to know you need to copy and paste it into a browser or click on it and open it at this step.

dylanhmorris: Done in [8b09e6e](https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/230/commits/8b09e6e8e8a0dde9e9cb6f4cae8c7060805419fc)

dylanhmorris:
Ready for re-review @kaitejohnson

batch/README.md
```diff
@@ -94,50 +106,150 @@ Note that if you close and reopen your Terminal window, you may need to reactiva
...
+1. Create or open the [`.bash_profile`](https://linuxopsys.com/dotfiles-in-linux-explained) file in your user home directory, i.e. the file located at
+```
+~/.bash_profile
+```
+
+3. Add the following line to your `.bash_profile`:
+```bash
+. azureconfig.sh
+```
+Note the `.`!
```
kaitejohnson:
```suggestion
Note the `. `!
```
Nitpicky but I think its important to make clear theres a space

batch/README.md
```diff
@@ -94,50 +106,150 @@ Note that if you close and reopen your Terminal window, you may need to reactiva
...
+az account show
+```
+- Be inside a Python virtual environment in which the dependencies specified in `batch/requirements.txt` have been installed.
+- Have appropriately environment variables. You can check this by trying to `echo` one to the Terminal:
+```bash
+echo $AZURE_BATCH_ACCOUNT
+```
+
+### Create or grab a configuration file
+We specify jobs using [YAML-formatted]() configuration files. We provide an `example_eval_config.yaml` in the `wastewater-input` Blob storage container within the `cfaazurebatchprd` Blob storage account. Download it to your local machine using the [Azure Storage Explorer] or via the command line, and save it the `wastewater-informed-covid-forecasting` directory under `input/config/eval/example_eval_config.yaml`.
```
kaitejohnson, line 165: I'm guessing we can't just commit this?

batch/README.md
```diff
@@ -94,50 +106,150 @@ Note that if you close and reopen your Terminal window, you may need to reactiva
...
+#### Model fitting
+Let's run `setup_job.py` to create a model fitting job and its constituent tasks. We'll name it `my-demo-fit-job` and have it run on the `wastewater-demo-pool` we just created. We'll use our local copy of the example configuration file (`example_eval_config.yaml`) and the corresponding copy of it Blob storage container `wastewater-input`.
+
+> [!CAUTION]
+> Make sure your local and remote config files are identical. Otherwise, the pipeline may error or behave unexpectedly. We may deduplicate the configs in a future refactor.
 
 ```bash
-pyhton3 batch/upload_data.py -g *.csv input/hosp_data wastewater-input
+python3 batch/setup_job.py input/config/eval/example_eval_config.yaml fit my-demo-fit-job wastewater-demo-pool
+```
```
kaitejohnson, line 212: Nice, so the change here is now in this step we are naming the job, but specifying the pool (which must already be created)?

kaitejohnson (dismissed):
This looks good to me, just two clarifying questions. Also would be good for @ztm5 to review!

batch/README.md
dylanhmorris: Good point (as it were).

batch/README.md
dylanhmorris: Yup!

kaitejohnson (dismissed):

batch/README.md
dylanhmorris: I wasn't sure so I didn't initially, but looking at it we can. Everything is either just configuration or is already public elsewhere (e.g. the gi and inf-to-hosp PMFs, which are in wwinference)

dylanhmorris: Done in db631a6

kaitejohnson (dismissed):
LGTM, thanks for adding the example config!

Issue #33, opened 2024-06-23
Deduplicate eval configs


kaitejohnson:
## Problem
config file needed from CLI and also on azure blob storage. Want to remove duplicative dependence on config because if they differ naming will be incorrectly mapped to whats running.

## Requirements
- [x] refactor R scripts to take in the arguments from the config that need to be passed to the script, parse them in python and pass the config in via the command line

## Context
Essentially, we want to rewrite setup_job.py to parse the config that it gets passed and to pass those as arguments in a loop to the command line R script. This way, we will never end up having a mismatch between what we pass to azure from the CLI and whats on blob storage.

Might end up being a lot of args though, and they all have to be in the correct order I think to work properly?

@damonbayer tagging you for awareness, I don't think this is an urgent task but part of the general updates to azure pipelining

PR #235, opened 2025-02-20
Simplify eval CLI

dylanhmorris:
## Goal 
Simplify and DRYify evaluation CLI in preparation for addressing #33  

## Changes
* Consolidates command line eval scripts into single script
* Updates `batch/setup_job.py` accordingly, DRYifies creation of tasks.
* Moves saved job stdout and stderr files into `logs` subdirectory of `raw_output`, making it easier to examine specific instances.

## Testing
* I have verified that the the job specified in `example_eval_config.yaml` successfully on Azure (both fitting and postprocessing)

## Notes
This is a complete PR, but it incorporates changes from #237. For ease/modularity, suggest reviewing and landing #237 before turning to this one.


kaitejohnson (commented):
This is looking pretty good, will wait til #237 is merged to fully review but thanks a lot for doing this @dylanhmorris 

To confirm, this just simplifies the workflow, it doesn't fix the problem that the `eval_config` passed into the command line has to be the same as the one on blob storage, correct?

dylanhmorris:
> To confirm, this just simplifies the workflow, it doesn't fix the problem that the eval_config passed into the command line has to be the same as the one on blob storage, correct?

Correct. The simplifications done here will make it easier to fix the eval config problem. I'm deferring the actual fix to another PR (#238) to keep things smaller/easier to review.

dylanhmorris:
With #237 landed this is now ready for review

.github/workflows/container-build-push.yaml
```diff
@@ -0,0 +1,63 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  IMAGE_NAME: renewalww
```
kaitejohnson: My understanding is that this is specified in the passed arguments to `setup_job.py` as described here https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/237#issuecomment-2675230339

Therefore, it won't always necessarily be `renewalww`. If we can't as a GH action take this as an argument, I don't think we should be allowing the user to specify that as an input. 

Wondering if there is a way for this action to grab that variable...

Containerfile
```diff
@@ -6,16 +6,9 @@ ADD ./.ContainerBuildRprofile /.Rprofile
 RUN Rscript -e "install.packages('pak')"
 RUN Rscript -e "pak::pkg_install('github::stan-dev/cmdstanr')"
 RUN Rscript -e "cmdstanr::install_cmdstan()"
-
+RUN Rscript -e "pak::pkg_install('github::cdcgov/ww-inference-model')"
+RUN Rscript -e "pak::pkg_install('argparser')"
```
kaitejohnson: Can you explain why we no longer need to compile the stan model in the container? Seeing that deleted

kaitejohnson: Is it just redundant because its already in the workflow for the fitting of the model?

run_eval.R
```diff
@@ -0,0 +1,65 @@
...
+    "eval_config_path",
+    help = "path to eval_config.yaml"
+  ) |>
+  add_argument(
+    "params_path",
+    help = "path to params.toml"
+  ) |>
+  add_argument(
+    "model",
+    help = "Model to fit (either 'ww' or 'hosp')"
```
kaitejohnson, line 30: In the `setup_job.py` you are passing in `"ww "` or `"hosp "`.  This appears that it would fail because this expects either "ww" or "hosp" with no spaces? 

Assuming you have tested this, curious how it hasn't failed and if the space is intentional?

run_eval.R
```diff
@@ -0,0 +1,65 @@
...
+
+run_name <- glue::glue("{parsed$job}_{parsed$model}")
+
+checkmate::assert_names(run_name,
+  subset.of = names(runner_functions)
+)
+
+job_runner_function <- runner_functions[[run_name]]
+
+message(glue::glue(
```
kaitejohnson, line 53: Can we add the model type here (wastewater or hospital admissions only?)

batch/setup_job.py
```diff
@@ -150,25 +151,24 @@ def main(
...
         this_scenario = "no_wastewater"
         task_name = (
-            f"{script_type}-{this_scenario}-"
+            f"{job_type}-{this_scenario}-"
             f"{this_forecast_date}-{this_location}"
         )
         base_call = (
             "/bin/sh -c '"
-            f"Rscript pipeline/command_line_eval_{script_type}_hosp.R "
+            f"Rscript run_eval.R "
```
kaitejohnson, line 132: To clarify for my own understanding, all these arguments need to be in the exact correct order, right? And so the user has to know to pass these in this exact order into the CLI?:
 ```
    config eval_config_file: str,
    job_type: str,
    job_id: str,
    pool_id: str,
    exclude_ww_model: bool
   ```
Or are these also from parsed arguments from the config? Could you edit the `batch/README.md` to describe how the workflow now differs?

kaitejohnson (commented):
Had a few questions

run_eval.R
dylanhmorris: Command line arguments are whitespace-separated by default.

The final call will look something like

```bash
Rscript run_eval.R 5 input/config/eval/example_eval_config.yaml input/params.toml ww fit 
```

In the definition of the call in `setup_job.py` we add those spaces are to separate the command line arguments from each other, so we don't get, e.g. 

```bash
Rscript run_eval.R 5input/config/eval/example_eval_config.yamlinput/params.tomlwwfit 
```

dylanhmorris: For the same reason, we already have spaces after the other arguments:

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/3f67dbd8d561a551d0bee032d3668c014eef3729/batch/setup_job.py#L134-L137

run_eval.R
dylanhmorris: the `run_name` combines both `job_type` and `model`. It will be one of `fit_ww`, `fit_hosp`, `postprocess_ww`, or `postprocess_hosp`. That said, probably clearer to split it up. Changed in 6dd1f6d

Containerfile
dylanhmorris: > Is it just redundant because its already in the workflow for the fitting of the model?

Exactly, and as currently set up the fitting workflow ignores any pre-compiled versions. The pre-compilation in the container is a vestige of an early attempt to streamline things by _not_ having individual jobs recompile the model, but that proved finicky and for most jobs fitting time is much longer than compile time.

batch/setup_job.py
dylanhmorris: First, a clarification: the arguments to the `main()` function need not be identical in meaning, name, or order to the arguments to whatever command line interface/argument parser we choose to define, though in general it's good practice to have those two things mirror each other.

For the command line interface,`eval_config_file`, `job_type` `job_id`, and `pool_id` are mandatory positional command line arguments. `--exclude-ww-model` is an optional flag argument (if not set, it is default `False`, i.e. we include the wastewater model).

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/3f67dbd8d561a551d0bee032d3668c014eef3729/batch/setup_job.py#L189-L227

These arguments then get passed to the `main` function here:
https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/3f67dbd8d561a551d0bee032d3668c014eef3729/batch/setup_job.py#L230

> And so the user has to know to pass these in this exact order into the CLI?

One advantage of using a helper to define a CLI (like Python `argparse` here or R `argarser` elsewhere in the project) is that you get an auto-documented command line interface. Here's what happens if you try to run `setup_job.py` with no arguments:

```bash
usage: setup_job.py [-h] [--container-image-name CONTAINER_IMAGE_NAME]
                    [--container-image-version CONTAINER_IMAGE_VERSION]
                    [--exclude-ww-model | --no-exclude-ww-model]
                    eval_config_file job_type job_id pool_id
setup_job.py: error: the following arguments are required: eval_config_file, job_type, job_id, pool_id
```

You can also run `setup_job.py -h` or `setup_job.py --help` and see a more verbose help message:

```bash
usage: setup_job.py [-h] [--container-image-name CONTAINER_IMAGE_NAME]
                    [--container-image-version CONTAINER_IMAGE_VERSION]
                    [--exclude-ww-model | --no-exclude-ww-model]
                    eval_config_file job_type job_id pool_id

Set up an Azure batch job from an evaluation configuration file.

positional arguments:
  eval_config_file      Path to a YAML-formatted configuration file
  job_type              Type of job to run (either `fit` or `postprocess`)
  job_id                Name for the Azure batch job
  pool_id               Name of the Azure batch pool on which to run the job

options:
  -h, --help            show this help message and exit
  --container-image-name CONTAINER_IMAGE_NAME
                        Name of the container to use for the job.
  --container-image-version CONTAINER_IMAGE_VERSION
                        Version of the container to use for the job.
  --exclude-ww-model, --no-exclude-ww-model
                        Exclude the wastewater model from fitting?
```

dylanhmorris: Made some updates to the batch README in 7d3b8ea

.github/workflows/container-build-push.yaml
dylanhmorris: Two points:
1. `setup_job.py` defaults to assuming the name of the container is `renewalww` if the user doesn't specify it. It's configurable only for future flexibility, not because I expect us often/ever to configure it differently in practice.
2. There's more than one way to build the container besides in CI. You could also build/tag/push it locally with a different name (e.g. `my-alternative-ww-container`), and then invoke use that alternate version with `setup_job.py {args} --container-image-name my-alternative-ww-container`

dylanhmorris: So my preference is to leave this as-is, but I'm happy to discuss further.

dylanhmorris:
Thanks, @kaitejohnson! Attempted to answer your questions and made some edits in response to your suggestions

run_eval.R
kaitejohnson: Hadn't realized that, ok this makes sense to me!

batch/setup_job.py
kaitejohnson: Yeah as long as we include an example of `setup_job.py -h` in the README.md then I think we're good -- I just know I woudld struggle to remember the order of all of these inputs now that they are not just accessed via the `batch_config.toml`.

.github/workflows/container-build-push.yaml
kaitejohnson: I think that's fine then, just might be worth noting that this only works if the container name is `renewalww` ?

batch/README.md
```diff
@@ -183,18 +183,18 @@ A job is a set of tasks. Each task (by default) gets handed to 1 "node" (virtual
...
 ```
-Rscript pipeline/command_line_eval_{script_type}_ww.R {config_index} input/config/eval/example_eval_config.yaml input/params.toml
+Rscript run_eval.R {config_index} input/config/eval/example_eval_config.yaml input/params.toml {model} {job_type}
 ```
 
-Invoking the command above performs either model fitting or model postprocessing for one of the forecasting problems specified in `example_eval_config.yaml`. A "forecasting problem" here means a forecast for a given location and date. Each forecasting problem has a corresponding `config_index` in the evaluation configuration `.yaml` file. For example, this command starts a fitting job for the 3rd entry in `example_eval_config.yaml`:
+Invoking the command above performs either model fitting or model postprocessing for one of the forecasting problems specified in `example_eval_config.yaml`. A "forecasting problem" here means a forecast for a given location and date. Each forecasting problem has a corresponding `config_index` in the evaluation configuration `.yaml` file. For example, this command starts a fitting job for the wastewater model with the 3rd entry in `example_eval_config.yaml`:
```
kaitejohnson, line 150: Should we say something about how the config is set up to list all combinations of locations and dates (e.g. each forecast problem)?

batch/README.md
```diff
@@ -291,3 +291,6 @@ make container_push
 ```
 
 Confirm that the container is now present in the registry by navigating to `portal.azure.com` and looking under `Resources > cfaprdbatchcr > services > repositories > renewalww`
+
+### Creating a configuration file
+The [`src/setup_eval.R`][../src/setup_eval.R] script ccan help you write properly formatted evaluation configuration YAML files. Remember to mirror config versions between your local `input/config/eval` directory and the one in your input Azure Blob storage container.
```
kaitejohnson:
```suggestion
The [`src/setup_eval.R`][../src/setup_eval.R] script can help you write properly formatted evaluation configuration YAML files. Remember to mirror config versions between your local `input/config/eval` directory and the one in your input Azure Blob storage container.
```

batch/README.md
```diff
@@ -205,7 +205,7 @@ To save you writing this all out by hand, `setup_job.py` loops over all the valu
 Let's run `setup_job.py` to create a model fitting job and its constituent tasks. We'll name it `my-demo-fit-job` and have it run on the `wastewater-demo-pool` we just created. We'll use our local copy of the example configuration file (`example_eval_config.yaml`) and the corresponding copy of it Blob storage container `wastewater-input`.
 
 > [!CAUTION]
-> Make sure your local and remote config files are identical. Otherwise, the pipeline may error or behave unexpectedly. We may deduplicate the configs in a future refactor.
+> Make sure your local and remote config files are identical. Otherwise, the pipeline may error or behave unexpectedly. We hope to deduplicate the configs in a future refactor.
```
kaitejohnson, line 167:
```suggestion
> Make sure your local and remote config files are identical. Otherwise, the pipeline may error or behave unexpectedly. We hope to deduplicate the configs in a future refactor. `setup_job.py` takes a number of command line arguments. For more details on each of them, run `setup_job.py --help` in the command line. 
```

kaitejohnson (commented):
Just a minor suggestion, otherwise looks good

batch/README.md
dylanhmorris: Done in 8c1db5a

batch/README.md
dylanhmorris: Addressed in 8c1db5a

.github/workflows/container-build-push.yaml
dylanhmorris: Addressed in 8c1db5a

dylanhmorris:
Should be ready now. Thanks for the reviews, @kaitejohnson!

kaitejohnson (approved):
Wondering if @ztm5 has any thoughts since he will can also test? I am not generally a huge fan of reviewing PRs without being able to run/test the code myself, but obviously ok in this case

PR #237, opened 2025-02-20
Build/push container in CI

### LINKED ISSUE #236: Build and push container in CI

dylanhmorris:
Closes #236 

For each PR, an associated version of the `renewalww` container gets built and pushed to the Azure CR. The version name is the name of the PR branch. For example, this PR's container version is `renewalww:dhm-use-container-action`.

The build from `prod` is tagged `latest`, so `renewalww:latest` should reflect the state of `prod`.

This is the approach taken in the [`pyrenew-hew`](https://github.com/cdcgov/pyrenew-hew) and  [`cfa-epinow2-pipeline`](https://github.com/CDCgov/cfa-epinow2-pipeline) repos, among others.


.containerignore
```diff
@@ -1,3 +1,11 @@
+Containerfile
+output
+scratch
+src
+lib
+lib64
+bin
+=======
```
dylanhmorris:
```suggestion
```

kaitejohnson (commented):
Exciting! I believe we previously used the `batch_config.toml` to point to the container name `renewalww:latest`. Where will this now be specified? I know you stated using `cfa-azure-tools` which meant that the `batch_config.toml` was redundant, so just remind me how the user would specify the name of the container to use in the jobs?

dylanhmorris:
It's via the `--container-image-name` and `--container-image-version` arguments to `batch/setup_job.py`. The former defaults to `renewalww` if not set and the latter defaults to `latest` if not set:

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/9d1eb1df5328aba134da3e7618a1601517cd98a3/batch/setup_job.py#L209-L221

kaitejohnson (approved):
LGTM!

PR #238, opened 2025-02-22
Deduplicate configs

### LINKED ISSUE #33: Deduplicate eval configs

dylanhmorris:
Closes #33


dylanhmorris:
Closing in favor of a new PR to address the same issue.

PR #240, opened 2025-02-23
DRYify `(load|save)_object` logic in `eval_fit.R` and `eval_postprocess.R`, move creation of evaluation datasets to postprocess

dylanhmorris:
These changes were made in the process of addressing #33 but are separable from it. I am making a separate PR to keep things more modular and easily reviewable.

## Changes
- Evaluation datasets are created (and saved to disk) in the `eval_postprocess.R` script rather than in the `eval_fit.R` script, since they are not used in the `eval_fit.R` script.
- The new `get_raw_output_suffix()` function DRYifies the computation of save suffixes for the `raw_output` folder.
- The new `to_rds_with_suffix()` function and calls to `purrr::partial` DRYify the creation of location/date/scenario/model-specific `save_object()` functions.
- In `to_rds_with_suffix()` we use `deparse(substitute(object)` to determine save names from object names in the R environment. This avoids `get()` calls and allow objects to be passed directly (as their names in the environment as strings). We also allow the user optionally to provide a fully custom save name (as in one that is completely distinct from the object name in the R environment), though we do not use that feature in practice.
- Unit tests for `to_rds_with_suffix()`

## Notes
Now that #235 has landed this is ready.


batch/setup_job.py
```diff
@@ -136,8 +136,8 @@ def main(
                 "input/params.toml "
                 "ww "
                 f"{job_type}"
-                f" > {log_dir}/stdout-{task_name}.txt "
-                f" 2> {log_dir}/stderr-{task_name}.txt"
+                f" > {log_dir}/{task_name}-stdout.txt "
```
kaitejohnson: I found it WAY easier to search these using `stdout` /`stderr` first, because you don't always know all the jobs that failed

kaitejohnson (commented):
Going to wait til 235 is landed to review line by line

batch/setup_job.py
dylanhmorris: It was half at the beginning (for ww) and half at the end (for hosp). Happy to standardize on "at the beginning", but note that now that they're in a separate `logs` folder they'll be easier to find anyway.

dylanhmorris: Argument for the end is that this way pairs from the same run will be together when sorted by name.

dylanhmorris:
@ztm5 @kaitejohnson now that #235 has landed this is ready.

dylanhmorris (commented):
Self-review, discussed ahead of time with @kaitejohnson

PR #241, opened 2025-02-25
Model-agnostic `eval_fit` function

dylanhmorris:
This DRYifies model fitting, which will also make it more straightforward to address #33

Deferring DRY of postprocessing to a separate PR.


dylanhmorris:
Self-review. Example eval runs end to end successfully.

PR #242, opened 2025-03-01
Update Containerfile

dylanhmorris:


dylanhmorris:
No longer needed as issue has been patched upstream

PR #243, opened 2025-03-04
Fix failing tests, add test check to CI

dylanhmorris:
Some tests were failing, and no tests were being run in CI due to the absence of a `testthat.R` file to flag things for R CMD check.

This removes some outdated tests and updates others, as well as fixing minor bugs in the main codebase that the tests caught.


dylanhmorris:
Self-review.

PR #244, opened 2025-03-04
Allow 'both' jobs in setup_job

dylanhmorris:
This allows a single job to handle both fitting and postprocessing. In that scheme, postprocessing tasks have their associated fitting tasks as dependencies, so they don't start until/unless the fitting jobs are complete. The manual first fit, then postprocess scheme remains supported.


dylanhmorris:
Self review: this works in test runs for both fitting and postprocessing.

PR #246, opened 2025-03-05
model agnostic postprocess functions

dylanhmorris:
Postprocessing is now handled by a single function regardless of whether it is for wastewater or admissions only model. This makes the pipeline simpler and makes it easier to address #238


Makefile
```diff
@@ -21,8 +21,11 @@ RMDIR = @rmdir
 RSCRIPT = Rscript --vanilla
 PYTHON = python3
 AZURE_CLI = az
-CONTAINER_ENGINE = docker
-CONTAINER_IMAGE_BUILD_CMD = $(CONTAINER_ENGINE) build
+DOCKER_COMMAND = podman
+CONTAINER_ENGINE = $(DOCKER_COMMAND)
```
kaitejohnson: Did I miss a separate PR that makes changes to the workflow and allows you to swap Docker or podman?

wweval/R/eval_postprocess.R
```diff
@@ -0,0 +1,548 @@
+#' Postprocess a successful eval fitting job.
+#'
```
kaitejohnson, line 2: Would be nice to add a description of what this generates precisely (e.g. saves quantities needed to evaluate forecast performance against observed data, etc. etc, and where these re saved). My understanding is that this doesn't use `save_table()` so this isn't putting anything into the nested file structure, just generating what I would call intermediates to facilitate saving the final outputs in the specified file structure in `save_table()`

wweval/R/eval_postprocess.R
```diff
@@ -0,0 +1,548 @@
...
+    output_dir = output_dir,
+    scenario = scenario,
+    forecast_date = forecast_date,
+    model_type = model,
+    location = location
+  )
+}
+
+#' Postprocess model for evaluation.
+#'
```
kaitejohnson, line 365: Would explain where things get saved in this (in nested file structure) and what gets saved, as well s the dependency on `postprocess_successful_fit()`

batch/setup_job.py
```diff
@@ -147,8 +146,8 @@ def add_task(
             "input/params.toml "
             f"{model} "
             f"{task_type}"
-            f" > {log_dir}/{task_name}-stdout.txt "
-            f" 2> {log_dir}/{task_name}-stderr.txt"
+            f" > {log_dir}/{task_id}-stdout.txt "
+            f" 2> {log_dir}/{task_id}-stderr.txt"
```
kaitejohnson, line 150: I think I am missing something, where in this workflow do you now call the function `eval_postprocess` instead of `eval_post_process()`. 

Previously all this was happening in `pipeline/command_line_eval_{ww/hosp}.R` I believe

kaitejohnson: Maybe I am missing something though, but yeah not seeing this here. Sorry really hard to follow without testing this stuff and now my memory is just getting hazier https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/dhm-model-agnostic-postprocess/command_line_eval.R

kaitejohnson (commented):
A bit confused on where these new functions are getting called

Makefile
dylanhmorris: Yes. https://github.com/CDCgov/wastewater-informed-covid-forecasting/pull/245

dylanhmorris: Really it shouldn't have shown up as changed at all. My mistake in not squash rebasing first. Fixed now.

wweval/R/eval_postprocess.R
dylanhmorris: > My understanding is that this doesn't use save_table() so this isn't putting anything into the nested file structure

it does use `save_table` etc. See e.g. 

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/b045e632d2d5aa2468ebaa8cef7d8b5e93289bd9/wweval/R/eval_postprocess.R#L92-L100

The main purpose of the `postprocess_successful_fit` function is readability. It wrap sall the commands that previously fell within this `else` clause in `eval_postprocess_ww`:

https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/b045e632d2d5aa2468ebaa8cef7d8b5e93289bd9/wweval/R/eval_postprocess.R#L500-L513

When all of what is now in the body of `postprocess_successful_fit` was within that `else { }`, it was hard to see where the else ended.

wweval/R/eval_postprocess.R
dylanhmorris: Made an attempt to improve/clarify docs in ff61c0a

batch/setup_job.py
dylanhmorris: Yes, the `commdn_line_eval.R` script was deprecated a while back and should have been removed a while ago (have now done that in 2da90ec). As of the old release , we were using `pipeline/command_line_eval_fit_ww.R`, `pipeline/command_line_eval_postprocess_ww.R`, etc (see [old README here](https://github.com/CDCgov/wastewater-informed-covid-forecasting/tree/dd56f059d2bdcb47794d74e2cc9ab2e7c4ef3a41/batch#set-up-a-specific-job)). In #235 I simplified the CLI so that all of those scripts can be replaced by a single `run_eval.R`.

The [current version of the readme](https://github.com/CDCgov/wastewater-informed-covid-forecasting/tree/prod/batch#set-up-a-specific-job) reflects this.

dylanhmorris: > do you now call the function `eval_postprocess()` instead of `eval_post_process()`.

Yes, that was a simple rename for consistency. In some places in the codebase "post-process" appeared as one word ("`postprocess`") and other places as two words ("`post_process`").

kaitejohnson: I see -- I missed the changes to that script. This makes sense to me.

kaitejohnson (approved):
This looks good to me presuming its been tested!

dylanhmorris:
> This looks good to me presuming its been tested!

Thanks! Yes, it has.

PR #249, opened 2025-03-08
wweval pkgdown site

dylanhmorris:
This PR adds the `wweval` pkgdown site as this repo's GH pages site, since the `cfaforecatrenewalww` package is defunct. This should improve repo navigability and result reproducibility. Also should be helpful internal reference for any remaining analyses.


kaitejohnson:
My thought on this is that we really don't need a pkgdown site for either of these, but since we had already made one for `cfaforecastrenewalww` this makes sense.

kaitejohnson (approved):

PR #250, opened 2025-03-10
Add script to pull latest data for eval

dylanhmorris:
This will allow us to evaluate one more forecast week.


kaitejohnson (approved):
This looks good to me, I think maybe this should just be a package function (e.g. in R?). Previously, all the vintages had just been saved using the old `cfaforecastrenewalww` using either covidcast/epidatr or NHSN API directly depending on if real-time or retrospectively. But would be good to have documentation on this function, and even an example. 

Here `output_dir` will need to be specified locally, and then this will have to be moved to the location of the vitnages in blob storage, correct?

dylanhmorris:
> This looks good to me, I think maybe this should just be a package function (e.g. in R?). Previously, all the vintages had just been saved using the old cfaforecastrenewalww using either covidcast/epidatr or NHSN API directly depending on if real-time or retrospectively. But would be good to have documentation on this function, and even an example.

Good idea, will implement. Done in d49a5c6

> Here output_dir will need to be specified locally, and then this will have to be moved to the location of the vintages in blob storage, correct?

Yes, though part of the idea is that if you have blob containers mounted locally you can just write straight to them. I did this, in fact.

PR #251, opened 2025-03-10
use only a local eval config, make postprocessing more robust and faster

### LINKED ISSUE #33: Deduplicate eval configs

dylanhmorris:
Closes #33 

Uses changes from #246, which should be landed first. (EDIT: it has).

## Other changes:
- Reorganizes postprocessing functions within `wweval` into a single `postprocess.R` file (previously: `post_process.R` and `eval_postprocess.R`)
- Changes wastewater data mapping for most recent available from `NULL` to an explicit "most recent" and updates tests accordingly
- Adjusts pulling of wastewater evaluation data to make it more robust to choice of evaluation date.
- Adds functionality to run jobs for individual locations (via a `--locations-only` flag), which is useful for debug reruns or postprocessing locations with large memory requirements on a larger VM.
- Refactors the wastewater quantile computation code to make it faster and more memory-efficient.


batch/setup_job.py
```diff
@@ -18,6 +18,7 @@ def main(
     exclude_ww_model: bool,
     container_image_name: str = "renewalww",
     container_image_version: str = "latest",
+    locations_only: list[str] = None,
```
kaitejohnson, line 21: This is nice (though somewhat redundant if the config can now easily be modified in one place). The way I previously did this was just writing a new smaller config and putting it on blob and locally/

batch/setup_job.py
```diff
@@ -87,9 +95,7 @@ def main(
             print(f"Job {job_id} already exists.")
 
     container_image = (
-        f"{creds.azure_container_registry_account}."
-        f"{creds.azure_container_registry_domain}/"
-        f"{container_image_name}:{container_image_version}"
+        f"ghcr.io/cdcgov/{container_image_name}:{container_image_version}"
```
kaitejohnson, line 98: Is this something new with the CR? I might have missed some changes, but trust that this is now working more smoothly (e.g. automated container updates upon pushes to main?)

batch/setup_job.py
```diff
@@ -141,13 +145,28 @@ def add_task(
...
+            f"--forecast-horizon {eval_spec['forecast_time']} "
+            f"--params-path input/params.toml "
+            f"--output-dir {eval_spec['output_dir']} "
+            f"--raw-output-dir {eval_spec['raw_output_dir']} "
+            f"--seed {eval_spec['seed']} "
+            f"--iter-sampling {eval_spec['iter_sampling']} "
+            f"--n-chains {eval_spec['n_chains']} "
+            f"--adapt-delta {eval_spec['adapt_delta']} "
+            f"--max-treedepth {eval_spec['max_treedepth']} "
+            f"--task-type {task_type} "
```
kaitejohnson, line 167: This doesn't change any of the user CLI arguments does it?

run_eval.R
```diff
@@ -9,51 +9,130 @@ runner_functions <- c(
...
+      "sampler adaptation phase."
+    )
+  ) |>
+  add_argument(
+    "--max-treedepth",
+    help = "Maximum tree depth for the No-U-Turn sampler."
+  ) |>
+  add_argument(
+    "--task-type",
+    help = "Task to run. One of 'fit' and 'postprocess'"
```
kaitejohnson, line 116: I think I'm just a little confused, are all of these arguments the user has to put into the command line?

src/setup_eval.R
```diff
@@ -18,7 +18,7 @@ write_eval_config(
     as.character(
       seq(
         from = lubridate::ymd("2023-10-16"),
-        to = lubridate::ymd("2024-03-11"),
+        to = lubridate::ymd("2024-04-29"),
```
kaitejohnson, line 21: We shouldn't be generating forecasts through 4-29 I don't think, wouldn't it just be 2 extra weeks so 3-25?

kaitejohnson (commented):
This looks overall good with the caveat that its hard to know the new flow without testing. 

Only concerns are the few questions I have and updates to the instructions for the user in the batch readme.

batch/setup_job.py
dylanhmorris: I like being able to use the same config for both as it enforces 1 config --> 1 set of outputs. That in turn helps ensure that other key settings are as expected and prevents duplication of files.

batch/setup_job.py
dylanhmorris: Yes, I have tested this and it works. I was asked to modify this by CDCGov admins, who wanted to reduce/eliminate use of self-hosted runners on Gov repos. Some in the weeds details follow if you're interested.

Previously, our container was built on a self-hosted runner and pushed to a private Azure Container Registry repo we own. But there's nothing sensitive in our container. The image contains only fully public code. So we can just build it on a github-hosted runner and deploy it to the (public) Github container registry. Azure Batch can pull containers from public registries; it doesn't have to work only with containers on a private ACR instance. So I've set that up and modified the batch readme accordingly.

batch/setup_job.py
dylanhmorris: It doesn't change the CLI for `setup_job.py`, but `run_eval.R` now expects more command line arguments (see below). The code here extracts the needed values from the provided eval_config yaml and passes them to `run_eval.R`

run_eval.R
dylanhmorris: A user does not typically call `run_eval.R` from their local command line (versus having it called in cloud jobs). If when they do, they do need to provide these args. 

It would be possible to configure `run_eval.R` to take a mix of arguments from a config file and command line arguments (with one overloading the other), but I'm disinclined to do that. Local runs of `run_eval.R` are in practice rare. I haven't found it too onerous in practice, and it makes it much easier to confirm that each batch task is running with the correct configuration (by inspecting the command line call).

src/setup_eval.R
dylanhmorris: I ran everything available to confirm everything worked (even with fits we don't need), but can revert for any final production runs.

dylanhmorris:
Thanks @kaitejohnson. LMK if this answers your questions. I have run everything in practice and it works. I have also updated the batch README/walkthrough; thanks for flagging it!

run_eval.R
kaitejohnson: This makes more sense now that I am following the flow from `setup_job.py` --> `run_eval.R`

src/setup_eval.R
kaitejohnson: Ok yeah -- would revert this for production runs because as is this will average over scores which will have different horizon lengths (since 4-29 will not be evaluated 4 weeks out)

kaitejohnson (approved):
This LGTM, thanks for the clarifications

PR #267, opened 2025-04-17
Fix local container build and instructions

dylanhmorris:


dylanhmorris:
I have confirmed the instructions work as written, so I am merging.

PR #270, opened 2025-04-23
Switch to `scoringutils` 2.1.0

dylanhmorris:
- Makes some simple changes to the codebase allowing us to use the latest version of `scoringutils`.
- Refactors some scoring functions to reduce complexity while preserving functionality (e.g. by replacing nested loop with a map).

### DIFF
--- a/wweval/DESCRIPTION
+++ b/wweval/DESCRIPTION
@@ -70,7 +70,7 @@ Imports:
     soql,
     tibble,
     zoo,
-    cmdstanr (>= 0.7.1),
+    cmdstanr (>= 0.9.0),
     forcats,
     ggridges,
     cowplot,
@@ -78,9 +78,9 @@ Imports:
     RColorBrewer,
     zoltr,
     wwinference,
-    scoringutils (>= 1.2.2),
     fs,
-    checkmate
+    checkmate,
+    scoringutils
 Suggests:
     httr,
     sessioninfo,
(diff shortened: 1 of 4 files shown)


wweval/DESCRIPTION
```diff
@@ -70,17 +70,17 @@ Imports:
     soql,
     tibble,
     zoo,
-    cmdstanr (>= 0.7.1),
+    cmdstanr (>= 0.9.0),
```
kaitejohnson, line 73: Have we checked to make sure this doesn't break anything?

kaitejohnson (commented):
This particular code will work fine but there are a number of functions that are used in the `targets_eval_postprocessing.R` (that you will rename but I haven't seen this yet) that expect the outputs of `scoringutilsv1.2.2` (these are mainly the calibration plots -- interval and qq plots). 

In particular to name a few spots there are:
- https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/42a0b026b957ed47b089cb7be5aa958d93c86a85/_targets_eval_postprocessing.R#L739
- https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/42a0b026b957ed47b089cb7be5aa958d93c86a85/_targets_eval_postprocessing.R#L821
- https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/42a0b026b957ed47b089cb7be5aa958d93c86a85/_targets_eval_postprocessing.R#L1023
- https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/42a0b026b957ed47b089cb7be5aa958d93c86a85/_targets_eval_postprocessing.R#L1032
- As well as I think needing to specify the attributes for the score based dataframes.

I started making a bunch of changes in this PR a while back and we decided to close it because it was a lot of changes and we were near the final stages.

I think it might be easier to stick with 1.2.2, with the only concern being the crps breakdown-- this might require the later version...

dylanhmorris:
Superceded by #271

PR #272, opened 2025-05-07
Save posterior samples of admissions as a parquet file during postprocessing

dylanhmorris:


dylanhmorris:
Will handle this in #271 instead.

PR #273, opened 2025-05-08
Replace loop accumulation in combine_outputs with a pmap approach

dylanhmorris:
This should make read in more robust and faster


wweval/R/combine_outputs.R
```diff
@@ -22,88 +22,94 @@
...
-          this_output <- readr::read_tsv(fp)
-          combined_output <- rbind(combined_output, this_output)
-        },
-        error = function(e) {}
-      )
+      output <- readr::read_tsv(
+        fp,
+        show_col_types = FALSE
+      ) |>
+        dplyr::mutate(success = TRUE)
```
kaitejohnson, line 70: I think you just want to test this and make sure that the entire read-in is falesafe ( I am assuming you did this). Otherwise this will get stuck at a missing file.

kaitejohnson (commented):
LGTM pending the failing safely

wweval/R/combine_outputs.R
dylanhmorris: Yes, the safety is handled by the `file.exists`. I removed the additional try-catch because I think it _should_ fail if a named `tsv` exists but is unreadable.

kaitejohnson: Yep, that makes sense to me!

PR #290, opened 2025-07-02
Compute and save posteriors of the differences between the forecasts

dylanhmorris:
Closes https://github.com/cdcent/cfa-forecast-renewal-ww/issues/804

Also scope-creep tweak to postprocess.R to make period categorization a little more robust in `NA` handling.


wweval/R/constants.R
```diff
@@ -0,0 +1,2 @@
+#' Difference in days between standard forecast dates
+#' and typical last hospital admission data date.
```
dylanhmorris:
```suggestion
```

_targets.R
```diff
@@ -57,7 +57,7 @@ upstream_targets <- list(
   ),
   tar_target(
     name = last_real_time_forecast_date,
-    command = lubridate::ymd("2024-04-29")
+    command = lubridate::ymd("2024-03-25")
```
kaitejohnson, line 60: Confirming (bc I can't remember) that this is the last date we now are using to evaluate?

kaitejohnson (approved):
Quick question, otherwise LGTM

_targets.R
dylanhmorris: Correct!

PR #293, opened 2025-07-10
Improve figure exporting, add remaining trendfit figures

### LINKED ISSUE #233: Do not display unscored dates on plots
2024-03-18 is not currently scored but shows up (blank) on most/all plots of scores through time (both lineplots and heatmaps)

dylanhmorris:
- Closes https://github.com/cdcent/cfa-forecast-renewal-ww/issues/811
- 
Helps with 
- https://github.com/cdcent/cfa-forecast-renewal-ww/issues/806
- https://github.com/cdcent/cfa-forecast-renewal-ww/issues/807

## Minor changes
Closes #233 

#291 should land first.


kaitejohnson:
Could we add one more figure which is diff(slope_ww, slop_hosp) vs CRPS ratio??

kaitejohnson (approved):
LGTM, see my one request for the slope_ww vs slope_hosp colored by relative CRPS? (which I know is similar to relative CRPS vs diff in hosp but just might be easier to reason about)

PR #328, opened 2025-12-09
Simplify get_convergence_df using summarise by and if_any

dylanhmorris:


dylanhmorris:
Have verified that this produces equivalent results to the old implementation.

PR #332, opened 2025-12-09
Vary slopes by site, not site-lab when inferring recent trends

dylanhmorris:
- Also switches to 0-indexed time from 1-indexed.


dylanhmorris:
Not planned.

PR #336, opened 2025-12-16
Add computation of additional mean score ratios of interest

dylanhmorris:


kaitejohnson (commented):
Shouldn't there also be one that's by location? Or did we not report that.

I also struggle a bit to follow this in that its using forecasttools. 

I thin you could do this just as easily and less computation with dplyr. Assuming the first argument is the full set of scores at the most granular level, you could just summarise by the stratification and then pivot wider, divide by the baseline, pivot back to longer, and find the max/min?

This is how it is done in each of the figures that present these distributions, so it may be better to be consistent.

dylanhmorris:
> Shouldn't there also be one that's by location? Or did we not report that.

We already had that. See https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/6a254f7690b37774a5c6326419ca760a859296da/_targets.R#L2342

> I also struggle a bit to follow this in that its using forecasttools.
>
> I thin you could do this just as easily and less computation with dplyr. Assuming the first argument is the full set of scores at the most granular level, you could just summarise by the stratification and then pivot wider, divide by the baseline, pivot back to longer, and find the max/min?

It's [a light wrapper of `scoringutils`'s model comparison functionality](https://cdcgov.github.io/forecasttools/reference/summarise_scores_with_baseline.html). I prefer using it to doing things via manual split-apply-combine both for DRY and because it leverages a well-tested package. `summarise_scores_with_baseline` [is also tested](https://github.com/CDCgov/forecasttools/blob/main/tests/testthat/test_summarise_scores_with_baseline.R).

> This is how it is done in each of the figures that present these distributions, so it may be better to be consistent.

I refactored those to use the approach here for the reason above.

kaitejohnson (approved):

PR #339, opened 2026-01-13
Adjust palette to make Hub models grey

dylanhmorris:
Also removes no longer present models from palette

Closes https://github.com/cdcent/cfa-forecast-renewal-ww/issues/859


kaitejohnson (approved):
LGTM!

PR #355, opened 2026-03-03
Update plot saving to save two formats by default

### LINKED ISSUE #354: Save figures from targets pipeline in multiple, configurable formats
1 raster (jpg) format for quick looks and inclusion in slides etc, 1 vector format (eps)

dylanhmorris:
Closes #354 

`format = "file"` specs removed for figure-saving targets because they now have two side effects (writing two files) rather than a single one. This is fine because nothing is downstream of any of those written files.

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -31,12 +31,12 @@ save_figure <- function(
     fig_name <- deparse(substitute(fig))
   }
   purrr::walk(formats, \(format) {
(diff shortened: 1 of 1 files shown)


dylanhmorris:
I have tested this out and it works as expected (by default, saves a 300 DPI jpg + an eps)

kaitejohnson (approved):
LGTM, thanks for adding this!

